PR #25, opened 2019-11-01
global: several bug fixes

ppanero:
- When executing the build in "semi-prod" the app and the base images needed to be explicitly requested (i.e. add ``--base`` and ``--app`` flags). It is better to have them by default and allow ``skip`` on them.

- There was a bug on the image name, it was containing the whole cwd path instead of just the project name. A new property was added.

- The destroy command was not removing the volumes. Kudos to @fenekku, for the reminder :)

### DIFF
--- a/invenio_cli/cli.py
+++ b/invenio_cli/cli.py
@@ -39,6 +39,8 @@ class InvenioCli(object):
         :param flavour: Flavour name.
         """
         self.cwd = None
+        self.flavour = None
+        self.name = None
         self.config = ConfigParser()
         self.config.read(CONFIG_FILENAME)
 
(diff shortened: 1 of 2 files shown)


invenio_cli/cli.py
```diff
@@ -107,10 +110,10 @@ def init(cli_obj):
 @click.pass_obj
 @click.option('--dev/--prod', default=True, is_flag=True,
               help='Which environment to build, it defaults to development')
-@click.option('--base', default=False, is_flag=True,
+@click.option('--base/--skip-base', default=True, is_flag=True,
```
fenekku: nice, didn't know about `/` in `click.option`

invenio_cli/utils.py
```diff
@@ -20,19 +20,19 @@ class DockerCompose(object):
     def create_containers(dev, cwd):
         """Create images according to the specified environment."""
         command = ['docker-compose',
-                   '-f', 'docker-compose.yml', 'up', '--no-start']
+                   '-f', 'docker-compose.full.yml', 'up', '--no-start']
         if dev:
-            command[2] = 'docker-compose.dev.yml'
+            command[2] = 'docker-compose.yml'
```
fenekku: :+1:

PR #31, opened 2019-11-05
docs: update README with new cli usage

fenekku:


README.rst
```diff
@@ -26,29 +26,45 @@
...
+
+Just like the above, except with ``--prod``:
+
+.. code-block:: console
+
+    # Initialize environment
+    $ invenio-cli --flavour=RDM init
+
+    # Build Docker images
+    $ invenio-cli build --prod --pre --lock
```
ppanero: Minor: Since `--pre` installs alpha releases shall we remove `--pre` for prod? (Just in the docs since we need it atm).

fenekku: Yes. Let's reinforce that prod should be with solid non-alpha libraries.

Issue #35, opened 2019-11-07
global: update NU copyright


ppanero:
It should be ``Copyright (C) 2019 Northwestern University.`` instead of ``# Copyright (C) 2019 Northwestern University, Galter Health Sciences Library & Learning Center.``

Issue #21, opened 2019-10-30
init: folder already exists


ppanero:
If the folder generated by cookiecutter already exsits, it thows an exception. It is not possible to control the flow (get the name of the folder) before cookicutter finishes. The exception should be controlled and presented to the user in a friendly way.

Issue #20, opened 2019-10-30
init: file already exists error


ppanero:
The init command creates a `.invenio` file in order to store some configuration. If the file already exists, it fails after the cookicutter process, it should check it beforehand.

PR #36, opened 2019-11-07
Several bug fixes

ppanero:
Changes:

- Change NU copyright string across the module.
- Modify config file location. The `.invinio` configuration file should be created inside the project folder. This would allow creation of multiple projects using the CLI while still being in the same parent directory. The caveat is that for later commands the user needs to go inside the project folder (which is somehow desired so it is explicit which project is building/running/destroying).
- Catch the exception if the folder already exists. Hopefully, the fail will be faster in upcoming cookicutter versions (see https://github.com/cookiecutter/cookiecutter/issues/821)

Closes https://github.com/inveniosoftware/invenio-cli/issues/35
Closes https://github.com/inveniosoftware/invenio-cli/issues/20
Closes https://github.com/inveniosoftware/invenio-cli/issues/21

### DIFF
--- a/invenio_cli/cli.py
+++ b/invenio_cli/cli.py
@@ -133,7 +133,7 @@ def build(cli_obj, base, app, pre, dev, lock):
         subprocess.call(['/bin/bash', 'scripts/bootstrap', '--dev'])
 
         print('Creating development services...')
-        DockerCompose.create_containers(dev=True)
+        DockerCompose.create_images(dev=True)
     else:
         if base:
             print('Building {flavour} base docker image...'.format(
@@ -157,7 +157,7 @@ def build(cli_obj, base, app, pre, dev, lock):
                     project_name=cli_obj.name)
             )
         print('Creating full services...')
-        DockerCompose.create_containers(dev=False)
+        DockerCompose.create_images(dev=False)
 
 
 @cli.command()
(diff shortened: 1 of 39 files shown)


fenekku:
Since there isn't a release with this commit in it. It should be fine to be merged whenever right? (i.e. we can remove "WIP" from title)

invenio_cli/cli.py
```diff
@@ -17,6 +16,7 @@
 
 import click
 import docker
+from pathlib import Path
```
fenekku: minor: `pathlib` is part of the standard library in Python 3. This one is a backport, because setup.py says this module supports Python 2.7, but I thought we were only supporting Python 2 for previously existing module and only supporting Python 3 for new ones? We could get away with placing that import in the 1st party section and not installing the `pathlib` package.

invenio_cli/utils.py
```diff
@@ -17,16 +16,16 @@ class DockerCompose(object):
     """Utility class to interact with docker-compose."""
 
     # FIXME: Change name to create_images
```
fenekku: minor: Feel free to sneak the changes for that `#FIXME` in.

invenio_cli/cli.py
ppanero: You are right, Py 2 support it out for new modules. removing.

fenekku:
Re-reviewed. T'is good.

Issue #27, opened 2019-11-01
migrate scripts to commands


ppanero:
related to: https://github.com/inveniosoftware/cookiecutter-invenio-rdm/issues/3

PR #43, opened 2019-11-19
migrate bash scripts into cli [dev mode]

ppanero:
Closes https://github.com/inveniosoftware/invenio-cli/issues/27


invenio_cli/helpers/scripts.py
```diff
@@ -0,0 +1,298 @@
...
+    if verbose:
+        click.secho(res.output)
+
+
+#############
+# BOOTSTRAP #
+#############
+
+
+def _bootstrap_dev(pre, loglevel, logfile, verbose):
```
fenekku: This seems to indicate that log semantics have been diluted and then need to be condensed back.

What if we capture these "log information" in a single data structure and pass *it* around? This information always goes together.

invenio_cli/helpers/scripts.py
```diff
@@ -0,0 +1,298 @@
...
+                project_name=app_name)
+        )
+    if app:
+        click.secho('Building applications docker image...', fg='green')
+        # docker build -t my-site:latest .
+        docker_helper.built_image(
+            dockerfile='Dockerfile',
+            tag='{project_name}:latest'.format(
+                project_name=app_name)
+        )
```
fenekku: We can build the base, but not the app? I think we should just allow rebuilding the two together to limit the cases we need to keep track of.

invenio_cli/helpers/scripts.py
```diff
@@ -0,0 +1,298 @@
...
+        _bootstrap_dev(pre, loglevel, logfile, verbose)
+    else:
+        _boostrap_prod(base, app, docker_helper, app_name)
+
+
+@with_appcontext
+def build_assets(verbose):
+    """Build assets."""
+    cli = create_cli()
+    runner = current_app.test_cli_runner()
```
fenekku: Is there another runner than the test_cli one? I fear what runner's settings may be implied.

invenio_cli/helpers/scripts.py
```diff
@@ -0,0 +1,298 @@
...
+def update_config(dev, loglevel, logfile):
+    """Update invenio.cfg configuration file."""
+    if dev:
+        # Create symbolic link
+        dst_path = _get_instance_path(loglevel, logfile)
+        src_file = os.path.abspath('invenio.cfg')
+        _force_symlink(src_file, Path(dst_path) / 'invenio.cfg')
+
+    else:
+        # Copy to container
```
fenekku:
```suggestion
        # TODO: Copy to container
```
was this meant as a TODO?

invenio_cli/helpers/scripts.py
```diff
@@ -0,0 +1,298 @@
...
+    signal.signal(signal.SIGINT, signal_handler)
+
+    docker_helper.start_containers()
+
+    # Open logging pipe
+    logpipe = LogPipe(loglevel, logfile) if verbose else subprocess.PIPE
+
+    click.secho("Starting celery worker...", fg="green")
+    worker = subprocess.Popen(['pipenv', 'run', 'celery', 'worker',
+                               '-A', 'invenio_app.celery'],
```
fenekku: Let's expand the `-A`

invenio_cli/helpers/scripts.py
```diff
@@ -0,0 +1,298 @@
...
+        pass
+
+
+def _get_instance_path(loglevel, logfile):
+    # Open logging pipe
+    logpipe = LogPipe(loglevel, logfile)
+
+    calculate_path = subprocess.Popen(
+        ['pipenv', 'run', 'invenio', 'shell', '--no-term-title',
+            '-c', '"print(app.instance_path)"'],
```
fenekku:
```suggestion
            '-c', '"print(app.instance_path, end='')"'],
```

This should remove the need to remove the `\n` . Why do we need to `decode('utf-8')?

invenio_cli/helpers/scripts.py
fenekku: Encapsulate logfile and verbose.

Open issue about loglevel and remove until we get to it.

invenio_cli/helpers/scripts.py
fenekku: always app

invenio_cli/helpers/scripts.py
ppanero: https://github.com/inveniosoftware/invenio-cli/issues/49

PR #44, opened 2019-11-19
migrate scripts to cli [prod mode]

### LINKED ISSUE #45: prod: update logo
Currently the volume content is not updated.

ppanero:
Requires https://github.com/inveniosoftware/invenio-cli/pull/43

Closes #45


invenio_cli/helpers/docker_compose.py
```diff
@@ -73,10 +77,14 @@ def start_containers(self):
     def stop_containers(self):
         """Stop currently running containers."""
         # Open logging pipe
```
fenekku: minor: can remove

invenio_cli/helpers/docker_compose.py
```diff
@@ -73,10 +77,14 @@ def start_containers(self):
...
+        command = ['docker-compose', '-f', 'docker-compose.full.yml', 'stop']
```
fenekku: minor: expansion of `-f` into `--file`

invenio_cli/helpers/docker_compose.py
```diff
@@ -95,3 +103,14 @@ def destroy_containers(self):
...
+    def copy(self, src_file, dst_path, container):
+        """Copy a file into the path of the specified container."""
+        container = self.docker_client.containers.get(container)
+
+        with tarfile.open('tmp.tar', "w") as tar:
+            tar.add(src_file, arcname=os.path.basename(src_file),
+                    recursive=False)
+        with open('tmp.tar', 'rb') as fin:
+            data = io.BytesIO(fin.read())
+            container.put_archive(dst_path, data)
```
fenekku: Looking at https://github.com/docker/docker-py/blob/845d31dbd280f6c28cdb8594c6ff9d9f4b038bea/tests/integration/api_container_test.py#L679 . We should be able to do this (w/o io):

```suggestion
            container.put_archive(dst_path, fin)
```

invenio_cli/helpers/scripts.py
```diff
@@ -154,7 +156,10 @@ def update_statics(dev, loglevel, logfile):
         shutil.copyfile(src_file, dst_path / 'logo.svg')
 
     else:
-        pass
+        # dst_path is a path inside the container.
+        dst_path = '/opt/invenio/var/instance/static/images'
+        app_name = app_name.replace("-", "")
```
fenekku: `app_name` is `project_name` here and `project_name` may contain spaces too. `project_shortname` has spaces replaced with  `-`. 

Why are dashes problematic?

invenio_cli/helpers/docker_compose.py
```diff
@@ -95,3 +103,14 @@ def destroy_containers(self):
 
         # Close logging pipe
         logpipe.close()
+
+    def copy(self, src_file, dst_path, container):
```
fenekku: With the class dealing with more than docker-compose and being passed as `docker_helper`, changing the name to `DockerHelper` might be clearer.

invenio_cli/helpers/scripts.py
```diff
@@ -285,14 +279,14 @@ def signal_handler(sig, frame):
...
     """Run server."""
     if dev:
         _server_dev(docker_helper, loglevel, logfile, verbose)
     else:
-        _server_prod()
+        click.secho("Starting docker containers. " +
+                    "It might take up to a minute.", fg="green")
+        click.secho("Use --stop to stop server.", fg="green")
+        docker_helper.start_containers()
+        time.sleep(60)
```
fenekku: Is this needed because otherwise the python (parent) script finishes executing before the child is ready and therefore the child gets cancelled?

invenio_cli/helpers/scripts.py
ppanero: My bad, on semantics. `project_name` is `os.path.basename(context)` so in the end is `project_shortname`.

What about `app_name` all over?

about dashes, is just the way docker creates the containers :/ it removes the dashes so we need to comply with it...

invenio_cli/helpers/docker_compose.py
ppanero: True :) Fixed.

invenio_cli/helpers/scripts.py
ppanero: Nope, this is because the docker contianers take a bit to be up, and I want to avoid the user going to the browser and seeing a "site not reachable" error just because it was not patient enough :D

invenio_cli/helpers/docker_helper.py
```diff
@@ -105,9 +108,29 @@ def destroy_containers(self):
...
+        if dc_version < DOCKER_COMPOSE_VERSION_DASH:
+            return re.sub(r'[^-_a-z0-9]', '', app_name)
+        else:
+            return app_name
+
+    def copy(self, src_file, dst_path, app_name):
         """Copy a file into the path of the specified container."""
-        container = self.docker_client.containers.get(container)
+        container_name = '{}_web-ui_1'.format(self._normalize_name(app_name))
+        container = self.docker_client.containers.get(container_name)
```
fenekku: Idle thought: was there a way to get container by ID (that we could *actually* use) to get around needing this?

There are minuscule things worth a comment at least:
- `dc_version < DOCKER_COMPOSE_VERSION_DASH` will fail once  `dc_version = '1.1XX.X'` - which won't realistically happen, but is worth putting a comment to assuage other developers that may be concerned by low-probability edge cases
- `web` introduces a dependency on the docker-compose.yml file. Which again is fine, as long as we keep that in mind and don't change the names in the docker-compose.yml file.

invenio_cli/helpers/scripts.py
```diff
@@ -158,8 +158,7 @@ def update_statics(dev, loglevel, logfile, docker_helper=None, app_name=None):
     else:
         # dst_path is a path inside the container.
         dst_path = '/opt/invenio/var/instance/static/images'
-        app_name = app_name.replace("-", "")
-        docker_helper.copy(src_file, dst_path, '{}_web-ui_1'.format(app_name))
+        docker_helper.copy(src_file, dst_path, app_name)
```
fenekku: passed app_name here is the right one I assume.

PR #50, opened 2019-11-21
demo: add command to create demo records and fix prod setup

ppanero:
Changes:

* Executes setup commands inside the web_api container for `prod` version.
* Differences between docker-compose lower and greater than 1.21 (special characters accepatance).
* Adds command for data population (WIP in prod version)

### DIFF
--- a/invenio_cli/cli.py
+++ b/invenio_cli/cli.py
@@ -342,7 +342,7 @@ def upgrade(log_level, verbose):
 @click.option('--verbose', default=False, is_flag=True, required=False,
               help='Verbose mode will show all logs in the console.')
 def demo(dev, log_level, verbose):
-    """Upgrades the current application to the specified newer version."""
+    """Populates instance with demo records."""
     # Create config object
     invenio_cli = InvenioCli(
         loglevel=LEVELS[log_level],
@@ -350,4 +350,5 @@ def demo(dev, log_level, verbose):
     )
     docker_compose = DockerHelper(dev=dev, loglevel=invenio_cli.loglevel,
                                   logfile=invenio_cli.logfile)
-    populate_demo_records(docker_compose, invenio_cli.verbose)
+    populate_demo_records(dev, docker_compose, invenio_cli.verbose,
+                          invenio_cli.project_shortname)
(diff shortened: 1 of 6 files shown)


invenio_cli/helpers/scripts.py
```diff
@@ -295,4 +295,31 @@ def server(dev=True, docker_helper=None, app_name='invenio-rdm',
     if dev:
         _server_dev(docker_helper, loglevel, logfile, verbose)
     else:
-        _server_prod()
+        click.secho("Starting docker contianers. " +
```
fenekku:
```suggestion
        click.secho("Starting docker containers. " +
```

invenio_cli/cli.py
```diff
@@ -332,3 +332,22 @@ def upgrade(log_level, verbose):
...
+
+@cli.command()
+@click.option('--dev/--prod', default=True, is_flag=True,
+              help='Which environment to build, it defaults to development')
+@click.option('--log-level', required=False, default='warning',
+              type=click.Choice(list(LEVELS.keys()), case_sensitive=False))
+@click.option('--verbose', default=False, is_flag=True, required=False,
+              help='Verbose mode will show all logs in the console.')
+def demo(dev, log_level, verbose):
+    """Upgrades the current application to the specified newer version."""
```
fenekku:
```suggestion
    """Populates instance with demo records."""
```

invenio_cli/helpers/scripts.py
```diff
@@ -233,6 +221,62 @@ def setup(dev=True,  force=False, docker_helper=None,
...
+        click.secho("Purging queues...", fg="green")
+        docker_helper.execute_cli_command(
+            project_shortname, 'invenio index queue init purge')
+
+    click.secho("Creating database...", fg="green")
+    docker_helper.execute_cli_command(
+        project_shortname, 'invenio db init create',)
+    click.secho("Creating indexes...", fg="green")
+    docker_helper.execute_cli_command(
+        project_shortname, 'invenio index init --force')
```
fenekku: Should this `--force` be here?

ppanero: True, the is a use case for it (adding new schemas) but is not our case.

PR #65, opened 2019-12-03
scripts: fix logging in dev mode

ppanero:
Changes:

* Some parts in the dev mode were not correctly fixed in the previous merge.

NOTE: This also points out the lack of tests... In January we should really push to have tests once we have the command naming fixed.


invenio_cli/helpers/scripts.py
```diff
@@ -314,7 +314,7 @@ def signal_handler(sig, frame):
     docker_helper.start_containers()
 
     # Open logging pipe
-    logpipe = LogPipe(log_config) if log_config.verbose else subprocess.PIPE
+    logpipe = subprocess.PIPE if log_config.verbose else LogPipe(log_config)
```
fenekku: This needed to be flipped?

Issue #66, opened 2019-12-03
Internal Server Error because build --dev should run commands via pipenv


fenekku:
Steps to reproduce error:

0- install invenio-cli globally (via sudo pip install invenio-cli or more recommended via [pipx](https://github.com/pipxproject/pipx))
1- `$ invenio-cli init --flavour=RDM`
2- `$ cd <your project short name>`
3- `$ invenio-cli build --pre --dev --verbose` 
Expected
4- assets are collected and built
Got
4-  Console output:

```console
$ invenio-cli build --pre --dev --verbose
Building RDM application...
Locking dependencies...
Bootstrapping server...
Checking that dependencies are locked...
Installing dependencies...
Collecting assets......
Usage: cli [OPTIONS] COMMAND [ARGS]...
Try "cli --help" for help.

Error: No such command "collect".

Building assets......
Usage: cli [OPTIONS] COMMAND [ARGS]...
Try "cli --help" for help.

Error: No such command "webpack".

Creating statics folder...
Statics directory already exsits...
Creating development services...
```

This is because 

```python
@with_appcontext
def build_assets(log_config):
    """Build assets."""
    cli = create_cli()
    runner = current_app.test_cli_runner()
    # Collect
    run_command(cli, runner, "collect -v", message="Collecting assets...",
                verbose=log_config.verbose)

    # Build using webpack
    run_command(cli, runner, "webpack buildall",
                message="Building assets...", verbose=log_config.verbose)
```

calls `run_command` directly rather than `subprocess.call(command, stdout=logpipe, stderr=logpipe)` where `command = ['pipenv', 'run', 'collect', '--verbose']` much like `_bootstrap_dev` .

Maybe this should be merged with #24 (and remove "good first ussue" from that one)

ppanero:
Previously (first release) the `--dev` version was using subprocess calls to run `pipenv run` calls. It was proven to be really slow (in the order of minutes compared to now).

Therefore, it was agreed to run inside an `application context`, which uses afterwards the command runner (see `run_command` function). The main drawback of this is the lack of support for having the CLI and the RDM application in different venvs (It was not considered).

Personally, I consider coming back to subprocess calls with `pipenv run` a step backwards. There is a clean way to solve this, which would be to somehow spawn an interpreter that loaded the current Invenio config and then run commands in there. That way we could profit of having the CLI separated and still load only once the application context (and avoid pipenv overhead for each operation).

[...]

Issue #24, opened 2019-10-31
build: allow assets regeneration without installing all deps


ppanero:
Allow regenerating assets (install/collect) without having to install all dependencies.

This could be added by adding a flag and splitting the boostrap script.

fenekku:
If nobody takes that one up soon, I would push for us to do it ourselves because it is a core improvement to the experience. It is already a drag ...  :D

PR #69, opened 2019-12-09
command: add assets command to build statics and/or assets independently

ppanero:
Changes:
- Adds an `assets` command.
- Allows collecting statics
- Allows building with webpack
- Allows both of the previous actions at the same time.

Caveats:
- The `webpack buidlall` part displays the full logs on the terminal. It is due to some internal part of it printing to stdout and now allowing control. It will be most likely be fixed by https://github.com/pallets/click/issues/737 and/or when we manage to fix https://github.com/inveniosoftware/invenio-cli/issues/66 using an interpreter.
- Not implemented for `--prod`. Since it requires a restart of uwsgi. I assume the workflow for this would be to `update` statics and rebuild the app container (not base) if not. All after having tested/modified in a `--dev` environment. (No ?)

Closes https://github.com/inveniosoftware/invenio-cli/issues/24

### DIFF
--- a/invenio_cli/helpers/scripts.py
+++ b/invenio_cli/helpers/scripts.py
@@ -62,7 +62,7 @@ def _bootstrap_dev(pre, log_config):
     logpipe.close()
 
     # Build assets
-    build_assets(log_config)
+    _build_local_assets(log_config)
 
     # Update static files
     update_statics(True, log_config)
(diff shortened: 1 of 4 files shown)


fenekku:
> Not implemented for --prod [...] workflow for this would be to update statics and rebuild the app container (not base) if not

Yes. This should be documented somewhere so it is clear: the workflow is edit your local work and run --dev. When ready generate --prod.

invenio_cli/helpers/scripts.py
```diff
@@ -101,17 +101,32 @@ def bootstrap(log_config, dev=True, pre=True,
...
-    run_command(cli, runner, "collect -v", message="Collecting assets...",
-                verbose=log_config.verbose)
+    if statics:
+        # Collect
+        run_command(cli, runner, "collect -v", message="Collecting assets...",
+                    verbose=log_config.verbose)
+    if webpack:
+        # Build using webpack
+        run_command(cli, runner, "webpack buildall",
+                    message="Building assets...", verbose=log_config.verbose)
```
fenekku: Shelved: It would be nice to break up, dependency (npm) installation and asset building too.

invenio_cli/helpers/scripts.py
```diff
@@ -101,17 +101,32 @@ def bootstrap(log_config, dev=True, pre=True,
...
+
+
+def _build_container_assets(log_config, statics=True, webpack=True):
+    """Build assets in the containers."""
+    # TODO: Check how to address this. Requires uwsgi restart.
+    pass
```
fenekku: Normal: Let's display a message that this is not something to do in prod / a full prod rebuild is what will take these changes into account. Important part here is to provide feedback to the user as opposed to no feedback.

ppanero: True :) fixed. Example output:
image

ppanero:
Created issues / addressed comments.

> Yes. This should be documented somewhere so it is clear: the workflow is edit your local work and run --dev. When ready generate --prod.

Agree, will start writing something. Maybe we can start creating more pages in the Wiki until we have a doc tool.

invenio_cli/helpers/scripts.py
fenekku: Nice!

PR #86, opened 2020-01-14
build: symlink templates/ for --local

### LINKED ISSUE #81: keep custom templates in sync with local / containers environment
Custom templates must somehow find their way in `<instance path>/templates/` . 

For `--local` environment, 2 potential avenues:
- copying
- symlink <---- Chosen

For `--containers` just copy the folder at image creation. (in view of new workflow)

Part of https://github.com/inveniosoftware/invenio-app-rdm/issues/6

fenekku:
- requires https://github.com/inveniosoftware/cookiecutter-invenio-rdm/pull/34 to be merged in
- `--containers` version is taken care of by the Dockerfile in cookiecutter (PR above)
- close #81

### DIFF
--- a/invenio_cli/helpers/scripts.py
+++ b/invenio_cli/helpers/scripts.py
@@ -70,14 +70,14 @@ def _bootstrap_local(pre, log_config):
     # Update configuration
     update_config(True, log_config)
 
+    symlink_templates_folder(log_config)
+
 
 def _boostrap_containers(docker_helper, project_shortname):
-    click.secho('Building applications docker image...', fg='green')
-    # docker build -t my-site:latest .
+    click.secho('Building application docker image...', fg='green')
     docker_helper.build_image(
         dockerfile='Dockerfile',
-        tag='{project_shortname}:latest'.format(
-            project_shortname=project_shortname)
+        tag='{}:latest'.format(project_shortname)
     )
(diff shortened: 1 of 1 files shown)


invenio_cli/helpers/scripts.py
```diff
@@ -122,28 +122,37 @@ def build_assets(local, statics, webpack, log_config):
...
     else:
         # Copy to container
         pass
 
 
+def symlink_templates_folder(log_config):
+    """Symlink the templates folder (only used for local development)."""
+    click.secho("Symlinking templates folder...", fg="green")
+    target_path = os.path.abspath('templates')
+    link_path = Path(_get_instance_path(log_config)) / 'templates/'
```
ppanero: minor: Is the `Path()` needed? I see up in `updated_config` function:

``` python
        link_path = _get_instance_path(log_config) / 'invenio.cfg'
```

If it is needed, it will be in both cases, if so, shall if be part of `_get_instance_path` function?

fenekku: Thanks for catching this! `Path` is needed. I made `_get_instance_path` return a `Path` (it even formats correctly: `'{}/data'.format(Path('/home/user')) == '/home/user/data'`)

PR #99, opened 2020-01-27
cli: create script for copying over instance theme folder

zzacharo:
Cli command to copy over instance `theme` folder generated here https://github.com/inveniosoftware/cookiecutter-invenio-rdm/pull/37. The configuration provided here is used here https://github.com/inveniosoftware/invenio-app-rdm/pull/59 .


invenio_cli/helpers/scripts.py
```diff
@@ -153,6 +154,26 @@ def symlink_templates_folder(log_config):
     _force_symlink(target_path, link_path)
 
 
+def update_theme(local, log_config, docker_helper=None):
```
ppanero: Moderate: Not sure if it might be worth since the code might completely change after @fenekku re-working of the CLI. But the code of `update_statics` and `update_theme` functions are quite similar. Could it be done generic just adding a one or two more parameters (e.g. file to search `src_file`).

zzacharo: @ppanero I totally agree with your proposal. I didn't spend more time on how this function would look like as I was thinking too that we will refactor the cli. I just duplicated the code I needed so that it would be easy to be refactored after :)

fenekku: My thinking is that I will put the theme part inside `update_statics` and probably not have an update container version. This is fine for now since like @zzacharo said, it will be easy to refactor.

Issue #100, opened 2020-01-29
helpers [cookiecutter]: make cookiecutter version configurable


ppanero:
To avoid hardcoding the cookiecutter version/branch we could provide a feature to load from config var also.

My idea is to:

- For monthly releases we still pin and use the correct version.
- If a user wants to use another version/tag (e.g. us for development) it can be done.

Issue #72, opened 2019-12-10
assets command: split npm and assets building


ppanero:
Currently, it is using `webpack buildall` from `flask-webpackext`. It would be nice to split the npm installation from the assets building (e.g. add npm flag?).

PR #102, opened 2020-01-31
init + build --local refactor

### LINKED ISSUE #41: tests: create tests for the result of each command.

### LINKED ISSUE #83: Installing invenio-cli through pipenv breaks --local commands
Easier to explain by listing the reproducing factors and then detailing what is happening:

Steps to reproduce:

1- install invenio-cli in a virtualenv and activate it
2- `(virtualenv)$ invenio-cli init --flavour=RDM --local`
3- `cd <project>`
4- `(virtualenv)$ invenio-cli build --pre --local`  (now have Pipfile and Pipfile.lock for the instance venv)
5- `(virtualenv)$ invenio-cli assets --local`
Expect:
6- assets to be collected and built
Get:
6- no assets collected or built

Running `(virtualenv)$ pipenv run invenio collect` and `(virtualenv)$ pipenv run invenio webpack buildall` does work.

This applies to any of the `--local` commands. 

~~My theory is it's because the created application (`create_cli`) is from the invenio-cli virtualenv context rather than the instance virtualenv context so none of the config is setup and there the commands are not operating on the right targets.~~

Core of the problem is that invenio-cli and the instance's installed dependencies are in different virtualenvs. This happens even if `PIPENV_IGNORE_VIRTUALENVS` is NOT set, because pipenv associates a Pipfile to a virtualenv: installing invenio-cli with pipenv will create a Pipfile that is different than the Pipfile `invenio-cli init` creates, so pipenv creates a new virtualenv to match *that* Pipfile everytime.

**WORKAROUND**
(extracted from comment below)

Say you have already installed invenio-cli through pipenv:

1- Run the init command to get the project laid out: `pipenv run invenio-cli init --flavour=RDM`
2- `cd <project folder>`
3- Install invenio-cli as part of the new Pipfile: pipenv install invenio-cli (check that it is there in the generated Pipfile)
4- Run the rest of the `invenio-cli --local` commands starting with: `pipenv run invenio-cli build --local --pre`

Because of this song and dance, we recommend these instructions for developing locally: https://github.com/ppanero/invenio-rdm-extension-demo/wiki

fenekku:
This is the first step in refactoring the cli:

- Only deals with init and build --local for now
- Illustrates relatively well how I will proceed with the rest of the commands: feedback now would be greatly appreciated to help orient things
- Removed the verbose flag for now to keep things a little more scoped. I do not mind at all adding it back later. 
- It adds tests: I decided to principally go the mocking way (there is temporary file creation for the cli config file only). I am experimenting here, if you've had prior experience testing similar things and have found productive approaches let me know. Good thing: tests are fast and cover a lot of things. Bad thing: confidence level is not as high given that some mocked calls expect specific input / return specific output in reality and that can only be known by running them for real. Closes #41 (the full refactor will)
- Sets #72 to be closed easily when working on update command: broken up buildall into create, install, build with `install` optional.
- Fixes #83 (at least for build --local , but will for rest)
- Note: this https://github.com/inveniosoftware/cookiecutter-invenio-rdm/pull/38 needs to be merged
- Will remove former code once all commands are dealt with
- It isn't rebased on @zzachero's work, but will be when that work is merged. 
- Wrapping cookiecutter in that CookiecutterWrapper will make #100 easier to address too... (I think)

I think that's all :D

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,6 @@ env:
   - REQUIREMENTS=devel
 
 python:
-  - "3.5"
   - "3.6"
 
 before_install:
--- a/invenio_cli/helpers/cli_config.py
+++ b/invenio_cli/helpers/cli_config.py
@@ -15,70 +15,55 @@ from pathlib import Path
 from .filesystem import get_created_files
 
 
-class InvenioCLIConfig(object):
+class CLIConfig(object):
     """Invenio-cli configuration."""
 
     CONFIG_FILENAME = '.invenio'
     CLI_SECTION = 'cli'
-    CLI_PROJECT_NAME = 'project_shortname'
-    CLI_FLAVOUR = 'flavour'
-    CLI_LOGFILE = 'logfile'
     COOKIECUTTER_SECTION = 'cookiecutter'
     FILES_SECTION = 'files'
 
-    def __init__(self, flavour=None, verbose=False):
-        """Initialize builder.
+    def __init__(self, fullpath=CONFIG_FILENAME, verbose=False):
+        """Constructor.
 
         :param flavour: Flavour name.
         """
-        self.flavour = None
-        self.project_shortname = None
-        self.log_config = None
         self.config = ConfigParser()
-        self.config.read(CONFIG_FILENAME)
+        self.fullpath = fullpath
+        self.config.read(fullpath)
 
-        # There is a .invenio config file
-        if os.path.isfile(CONFIG_FILENAME):
-            try:
-                self.flavour = self.config[CLI_SECTION][CLI_FLAVOUR]
-                self.project_shortname = \
-                    self.config[CLI_SECTION][CLI_PROJECT_NAME]
-                self.log_config = LoggingConfig(
-                    logfile=self.config[CLI_SECTION][CLI_LOGFILE],
-                    verbose=verbose
-                )
-            except KeyError:
-                logging.error(
-                    '{0}, {1} or {2} not configured in CLI section'.format(
-                        CLI_PROJECT_NAME, CLI_LOGFILE, CLI_FLAVOUR
-                    ))
-                exit(1)
-        elif flavour:
-            # There is no .invenio file but the flavour was provided via CLI
-            self.flavour = flavour
-        else:
-            # No value for flavour in .invenio nor CLI
-            logging.error('No flavour specified.')
-            exit(1)
+    def get_project_dir(self):
+        """Returns path to project directory."""
+        return Path(self.config[CLIConfig.CLI_SECTION]['project_dir'])
 
-    def read(self, fullpath):
-        """Read invenio-cli config file."""
-        pass
+    def get_instance_path(self):
+        """Returns path to application instance directory."""
+        return Path(self.config[CLIConfig.CLI_SECTION]['instance_path'])
+
+    def update_instance_path(self, new_instance_path):
+        """Updates path to application instance directory."""
+        self.config[CLIConfig.CLI_SECTION]['instance_path'] = \
+            str(new_instance_path)
+
+        with open(self.fullpath, 'w') as configfile:
+            self.config.write(configfile)
 
     @classmethod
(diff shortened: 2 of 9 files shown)


.travis.yml
```diff
@@ -28,7 +28,6 @@ env:
   - REQUIREMENTS=devel
 
 python:
-  - "3.5"
   - "3.6"
```
ppanero: Comment: We have actually discover (kudos to @zzacharo that its 3.6.2 the minimum). But I guess 3.6 is fine there. We can be more specific in end-user docs

invenio_cli/helpers/commands.py
```diff
@@ -0,0 +1,135 @@
...
+        # using it for a different purpose then intended but very useful
+        dir_util.copy_tree(src_dir, dst_dir)
+
+        # TODO: Uncomment when rebased on @zzachero work
+        # assets = 'assets'
+        # src_dir = self.cli_config.get_project_dir() / assets
+        # src_dir = str(src_dir)
+        # dst_dir = self.cli_config.get_instance_path() / assets
+        # dst_dir = str(dst_dir)
+        # dir_util.copy_tree(src_dir, dst_dir)
```
zzacharo: Just thinking that if people call  the `build` command again while everything is already built then the `copy_tree` command will fail as it requires `dst` folder not to exist. Also I think that the instance assets should be linked or we need to manually copy over in every change. This could fine for now but I am just saying that in some point we should integrate this behaviour :)

.travis.yml
fenekku: > We can be more specific in end-user docs

Ok. Travis is running Python 3.6.7 on Ubuntu 16.04.

invenio_cli/helpers/commands.py
fenekku: > Just thinking that if people call the build command again while everything is already built then the copy_tree command will fail as it requires dst folder not to exist.

Ah yes I faced that issue with `shutil.copytree` too. That's why I went with `distutils.dir_util.copy_tree` here. It doesn't have that problem.

+1 for the symlinking approach in the future. We are putting it on the backburner for now.

ppanero:
@fenekku one thing I just notice, before merging could this commit:
```
cookiecutter: set to master until release
```

Be squashed? I usually just change that in local but dont commit it. Otherwise we have it every release and its a bit "ugly" OCD point of view :P However https://github.com/inveniosoftware/invenio-cli/issues/100 this would fix it hehe it would allow us just to set and envvar to work with master.

fenekku:
@ppanero yeah... agreed it's not great: because it is annoying to go back and forth between `master` and `<release>`, but on the other hand, the changes in cookiecutter `master` *are* needed to make this work on the command line directly... 

(to make sure I understand, is it only the commit message that would like to remove or is it the change introduced?)

What are others thoughts on this? @zzacharo @lnielsen Whatever is the clearest / most effective for others I am fine with.

invenio_cli/helpers/cli_config.py
```diff
@@ -0,0 +1,82 @@
...
+
+        :param project_dir: Folder to write the config file into
+        :param flavour: 'RDM' or 'ILS'
+        :param replay: dict of cookiecutter replay
+        :return: full path to config file
+        """
+        config_parser = ConfigParser()
+
+        # Internal to Invenio-cli section
+        config_parser[cls.CLI_SECTION] = {}
```
zzacharo: @fenekku I think here you miss the `project_shortname` information :) I tried to run the cli locally but when it finishes the build and I am trying to start the server it requires to read the `project_shortname` from `cli` section in `.invenio.cfg` but it's not there. Do I miss something? Could you try it too and see if ti happens for you too?

fenekku: Thanks for noticing this. Yes I explicitly removed it, because I will be using the value in the "cookiecutter" section for that instead (rather than saving it twice in the file). I will add it back in this PR and then remove it, once I get to the running part.

PR #103, opened 2020-02-05
symlink assets folder

zzacharo:


invenio_cli/helpers/commands.py
```diff
@@ -34,6 +35,10 @@ def build(self, pre, lock):
         """Delegates to build of environment commands."""
         return self.environment.build(pre, lock)
 
+    def __getattr__(self, name, **kwargs):
+        """Delegate commands according to environment."""
+        return getattr(self.environment, name, **kwargs)
+
```
fenekku: Neat! I had just changed to a classmethod `create` rather than an `__init__` to return the 2 different classes and stop repeating method signatures, but I wasn't too happy about how it didn't connect the 2 implementations. I will go that way.

We might as well get rid of the `build` method above and make sure it is always called with named arguments.

invenio_cli/helpers/cookiecutter_wrapper.py
```diff
@@ -41,11 +41,11 @@ def repository(self):
         if self.flavour.upper() == 'RDM':
             self.template_name = 'cookiecutter-invenio-rdm'
             repo = {
-                'template': 'https://github.com/inveniosoftware/' +
+                'template': 'https://github.com/zzacharo/' +
                             '{}.git'.format(self.template_name),
                 # set to cookiecutter release version
                 # reset to master in development
-                'checkout': 'master'
+                'checkout': 'react-templates'
```
fenekku: I think you didn't mean to commit that file :)

zzacharo: oups :P

zzacharo:
@fenekku I will add a test, remove the `cookiecutter_wrapper.py` leftover and then push again :) I don't want to make more changes so you can focus on the refactoring part without any conflicts. After you can integrate these changes as you find more suitable in the overall redesign. What do you think?

PR #104, opened 2020-02-05
services: start+configure containers (--local)

fenekku:
Combines the idea of starting the services with their configuration: there won't be a need to run `invenio-cli setup`


fenekku:
Don't know why travis is yellow here given that it's passing here: https://travis-ci.org/inveniosoftware/invenio-cli/builds/646935485

fenekku:
Closed via #105

PR #106, opened 2020-02-11
cli: refactor commands (local+containers)

### LINKED ISSUE #49: flask command runner
Investigate if there is any other option to run the cli commands from `invenio` from within `invenio-cli` code. Currently we are using `current_app.test_cli_runner()` but it is meant for testing.

### LINKED ISSUE #55: global: change invenio-cli workflow (commands)
Text by @fenekku 

After some thoughts:

It actually might be that we have been thinking about this incorrectly: there shouldn't be 2 parallel contexts for which each command is implemented. Maybe it's more a sequence of commands and `--prod` is really just the last one or two. Something like this:
 
```
+------------------+     +---------------------+   +-------------------+      +--------------------+         +--------------------------+       +---------------------------------+
|                  |     |                     |   |                   |      |                    |         |                          |       |                                 |
| invenio-cli init +---> | invenio-cli service +-> | invenio-cli setup +----> | invenio-cli server +-------> | invenio-cli containerize +-----> | invenio-cli server --containers |
|                  |     |                     |   |                   |      |                    |         |                          |       |                                 |
+------------------+     +---------------------+   +-------------------+      +--------------------+         +--------------------------+       +---------------------------------+

```

Then being clear about the workflow will help users understand the 2 main use cases: 

1- I want to just see what InvenioRDM gives me out of the box: I go through `invenio-cli init`, `invenio-cli containerize` and `invenio-cli server --containers` so I can finally see the instance running with no configuration on my side. 

2- I am ready to start configuring and iterating: I do the first commands only and in particular the first `invenio-cli server` (precise names are not important in my explanation). This way my development flow is faster and I can see my changes but at the cost of more installation locally (might need to have header files and so on...).

This would make the usage of the cli more clear, remove the need for us to implement weird / necessarily slow commands and provide a better experience overall. What do you think @ppanero ? (Not saying we should switch now either if we decide to go this way)

### LINKED ISSUE #56: flask commands disambiguation
Flask commands from `invenio` get shown when using `invenio-cli --help`.

### LINKED ISSUE #58: handle exec commands status code
Modify behavior to take into account if a command fails.

### LINKED ISSUE #98: server: python autoreload not working on --local
Needs to default with env var `FLAKS_DEBUG=1` adn `FLASK_ENV`=development. One of them might suffice. To be tested.

### LINKED ISSUE #101: dependencies: fialing due to redis pin
``` python
Traceback (most recent call last):
  File "/home/ppanero/.virtualenvs/zach/lib/python3.6/site-packages/pkg_resources/__init__.py", line 584, in _build_master
    ws.require(__requires__)
  File "/home/ppanero/.virtualenvs/zach/lib/python3.6/site-packages/pkg_resources/__init__.py", line 901, in require
    needed = self.resolve(parse_requirements(requirements))
  File "/home/ppanero/.virtualenvs/zach/lib/python3.6/site-packages/pkg_resources/__init__.py", line 792, in resolve
    raise VersionConflict(dist, req).with_context(dependent_req)
pkg_resources.ContextualVersionConflict: (redis 3.4.0 (/home/ppanero/.virtualenvs/zach/lib/python3.6/site-packages), Requirement.parse('redis<3.4.0,>=3.3.11'), {'invenio-cli'})

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/ppanero/.virtualenvs/zach/bin/invenio-cli", line 6, in <module>
    from pkg_resources import load_entry_point
  File "/home/ppanero/.virtualenvs/zach/lib/python3.6/site-packages/pkg_resources/__init__.py", line 3252, in <module>
    @_call_aside
  File "/home/ppanero/.virtualenvs/zach/lib/python3.6/site-packages/pkg_resources/__init__.py", line 3236, in _call_aside
    f(*args, **kwargs)
  File "/home/ppanero/.virtualenvs/zach/lib/python3.6/site-packages/pkg_resources/__init__.py", line 3265, in _initialize_master_working_set
    working_set = WorkingSet._build_master()
  File "/home/ppanero/.virtualenvs/zach/lib/python3.6/site-packages/pkg_resources/__init__.py", line 586, in _build_master
    return cls._build_from_requirements(__requires__)
  File "/home/ppanero/.virtualenvs/zach/lib/python3.6/site-packages/pkg_resources/__init__.py", line 599, in _build_from_requirements
    dists = ws.resolve(reqs, Environment())
  File "/home/ppanero/.virtualenvs/zach/lib/python3.6/site-packages/pkg_resources/__init__.py", line 787, in resolve
    raise DistributionNotFound(req, requirers)
pkg_resources.DistributionNotFound: The 'redis<3.4.0,>=3.3.11' distribution was not found and is required by invenio-cli

```

Redis gets installed with `<3.4.0` in `invenio-cli`. Then upon `invenio-cli build` the installation happens according to the Pipfile where the pin of redis is only `>=2.10` or similar (coming form several invenio packages not invenio-app-rdm per se.

This issue has already happened with other libraries. But I think is going to be fixed when `invenio-cli` and the `invenio` commands/envs are separated? 

@fenekku could you confirm this (my assumption of it getting fixed). For now im just pinning in local, can provide a PR if its not the case of it getting fixed.

@zzacharo be aware of this when testing ;)

fenekku:
This changes the CLI commands (interface), changes some of the
code structure, adds more tests. Closes a series of issues.

demo, destroy and local to container workflow still need to be done.
(in addition to other cli issues)

Closes #55
Closes  #101 
Closes #58
Closes #56  
Closes #49 
Closes #98


README.rst
```diff
@@ -31,58 +31,64 @@ Installation
...
 
     # Initialize environment and cd into <created folder>
-    (<custom virtualenv>)$ invenio-cli init --flavour=RDM
-    (<custom virtualenv>)$ cd <created folder>
+    $ invenio-cli init --flavour=RDM
+    $ cd <created folder>
 
-    # Install python dependencies, link/copy assets + statics, install js
+    # Install locally
+    # install python dependencies, link/copy assets + statics, install js
```
ppanero: Doubt: assets + statics first they are copied/linked and then built? It's not all done in one command right? is like create + build?

README.rst
```diff
@@ -31,58 +31,64 @@ Installation
...
-    (<custom virtualenv>)$ invenio-cli build --local --pre
+    $ invenio-cli install
 
     # Start and setup services (database, Elasticsearch, Redis, queue)
-    (<custom virtualenv>)$ invenio-cli services --local
+    $ invenio-cli services
 
     # Optional: add demo data
-    (<custom virtualenv>)$ invenio-cli demo --local
+    $ invenio-cli demo --local
```
ppanero: Hope: I don'r really see how, but is there a way or getting rid of the `--local`/`--containers` for the demo command? We might even not consider it? and just say "If you want to use it you can do so through `invenio` (rdm-records)"?

README.rst
```diff
@@ -31,58 +31,64 @@ Installation
...
-    # Build Docker images
-    (<custom virtualenv>)$ invenio-cli build --containers --pre --verbose
+    # Optional: add demo data
+    $ invenio-cli demo --containers
 
-    # Setup services (database, Elasticsearch, Redis, queue)
-    (<custom virtualenv>)$ invenio-cli setup --containers --verbose
+    # After updating statics or code, if you do not need to re-install JS
+    # dependencies which can take time
+    $ invenio-cli containerize --no-install-js
```
ppanero: minor: For other parts we were using "skip" like "--skip-lock" in the *build* command (now *install*). So either we switch to `--no-lock` or `--skip-js` here? WDYT?

setup.py
```diff
@@ -83,26 +83,15 @@
         ],
         'flask.commands': [
```
ppanero: Moderate: Since we are switching out of "app-context" and removing old commands. We can already move back to console scripts:

```
        'console_scripts': [
            'invenio-cli = invenio_cli.cli:cli',
        ],
```

Instead of `flask.commands`. Would close https://github.com/inveniosoftware/invenio-cli/issues/56.

invenio_cli/cli.py
```diff
@@ -107,68 +53,51 @@ def init(flavour):
...
-    invenio_cli = InvenioCli(verbose=verbose)
-
-    click.secho('Setting up environment for {flavour} application...'
-                .format(flavour=invenio_cli.flavour), fg='green')
-
-    # Initialize docker client
-    docker_helper = DockerHelper(local=local,
-                                 log_config=invenio_cli.log_config)
+              help='Force recreation of db tables, ES indices, queues...')
+@click.option('--no-install-js', default=False, is_flag=True,
```
ppanero: minor: same `skip` vs `no...`

invenio_cli/cli.py
```diff
@@ -107,68 +53,51 @@ def init(flavour):
...
+    commands = LocalCommands(cli_config)
+    commands.install(pre=pre, lock=lock)
 
 
 @cli.command()
-@click.option('--local/--containers', default=True, is_flag=True,
-              help='Which environment to build, it defaults to local')
 @click.option('--force', default=False, is_flag=True,
               help='Force recreation of db tables, ES indices, queues...')
 def services(local, force):
```
ppanero: minor: I understood from the readme and what we talked on Gitter, that services won't need `--local`/`--containers`?

invenio_cli/cli.py
```diff
@@ -107,68 +53,51 @@ def init(flavour):
...
     """Starts services and ensures they are setup (DB, ES, queue, etc.).
 
-    TODO: Forward any extra argument to docker-compose.
+    --force destroys and resets services
     """
     cli_config = CLIConfig()
     commands = Commands(cli_config, local)
```
ppanero: minor: if the above comment is correct, this does not need `local` as parameter.

invenio_cli/cli.py
```diff
@@ -107,68 +53,51 @@ def init(flavour):
...
+    """Setup and run all containers (docker-compose.full.yml).
 
-    scripts_setup(local=local, force=force, stop_containers=stop_containers,
-                  docker_helper=docker_helper,
-                  project_shortname=invenio_cli.project_shortname,
-                  log_config=invenio_cli.log_config)
+    Think of it as a production compilation build + running.
+    """
+    cli_config = CLIConfig()
+    commands = ContainerizedCommands(
```
ppanero: moderate: There is a `Commands` factory, which already does this for you since you are passing `local=False`. so I think 

``` python
commands = Commands(cli_config, local=False)
```

 Would this suffice? or is there a reason why it is explicitly created this way?

invenio_cli/helpers/commands.py
```diff
@@ -157,10 +162,14 @@ def services(self, force):
...
         NOTE: We use check=True to mimic set -e from original setup script
               i.e. if a command fails, an exception is raised
+
+        A check is done to see if one-time setup has been executed before
+        and confirmation saved in invenio-cli config file.
         """
         self._ensure_containers_running()
 
         if force:
+            self.cli_config.update_services_setup(False)
```
ppanero: moderate: I think having this set is a really good idea. Nonetheless, I see a potential issue:

a) It's marked as set, an error occurs. Half are set half are not.
b) It's marked as not set, an error occurs. Half are set half are not.

In both cases retrying would case error in some of them... (would happen anyway if we dont have this flag, so having it is already an improvement to avoid setting if correct. So I think it would be an issue for the future (i.e. to be treated separately)

Conclusion: Should we run this after the setup commands? So we only have "set" if it's for sure set correctly? I think is better to have an error for "Some were set but it fails" rather than "They didnt got set but it said it is". I hope I explained myself correct, sorry if it's confusing :sweat_smile:

invenio_cli/helpers/commands.py
```diff
@@ -182,11 +191,16 @@ def services(self, force):
...
+        # Without the self.cli_config.get_services_setup() check
+        # this doesn't throw an error on re-runs
         command = ['pipenv', 'run', 'invenio', 'db', 'init', 'create']
         subprocess.run(command, check=True)
 
-        # If re-run, this throws an error
+        # Without the self.cli_config.get_services_setup() check
+        # this throws an error on re-runs
         # TODO: invenio-files-rest should make it idempotent
         command = [
```
ppanero: minor: not for now, but since we will incorporate S3 this might change to a factory of location commands or so? But highly shelvable :)

invenio_cli/helpers/cli_config.py
```diff
@@ -48,6 +48,22 @@ def update_instance_path(self, new_instance_path):
         with open(self.fullpath, 'w') as configfile:
             self.config.write(configfile)
 
+    def get_services_setup(self):
+        """Returns bool whether services have been setup or not."""
+        return self.config.getboolean(CLIConfig.CLI_SECTION, 'services_setup')
+
+    def update_services_setup(self, is_setup):
+        """Updates path to application instance directory."""
+        assert isinstance(is_setup, bool)
```
ppanero: doubt: in which case it would not be a boolean? I am just not a fan of asserts in the code

invenio_cli/helpers/commands.py
```diff
@@ -243,13 +262,117 @@ def signal_handler(sig, frame):
...
+        self.docker_helper = docker_helper
+
+    def update_statics_and_assets(self, install):
+        """Update application containers' statics and assets."""
+        project_shortname = self.cli_config.get_project_shortname()
+
+        click.secho('Collecting statics and assets...', fg='green')
+        # Collect into statics/ folder
+        self.docker_helper.execute_cli_command(
+            project_shortname, 'invenio collect --verbose')
```
ppanero: minor: Do we need verbose here as part of the command?

README.rst
fenekku: Right, it does create + install + build behind the scenes.
 
1- the invenio modules' statics/assets are collected first which creates/fills the `assets/` and `static/` folders in the instance path (collect and webpack create). 
2- Then the js dependencies are installed optionally, because we don't always want to call `npm install` which is slow, especially if we just changed some local file. See #72 
3- Then the *local assets* are copied / symlinked in the instance path, so that *all* the assets are in the same location
4- finally webpack build is run, which will build all the assets because they are now all present in the same `<instance_path>/assets/` folder.

README.rst
fenekku: On the question of having `--local/--containers`, It might be the case that running `invenio-cli demo` (without any option) would be enough to cover both cases, since they could potentially use the same DB container/volume. 

On the question of having `invenio-cli demo` at all,  it is quite convenient for the marketing case of: "just run these commands and you will see search results and records on your local installation". If we remove it, we have to tell people to run `docker exec <name of your UI container> /bin/bash -c "invenio demo"` which might be much just to fullfill the marketing case.

README.rst
fenekku: Oh interesting, I didn't notice that parallel: the `--skip-lock` is because that's what pipenv uses (we don't want to deviate from that and confuse people). I don't know that there is an expectation for `--no-install-js` to have a matching `--no-lock`... I could be swayed if other people think so.

I have noticed something related to that while trying out the workflow though:

- `invenio-cli containerize` should default to installing the JS dependencies, because that is the spirit of the `containerize` command: rebuild everything (except for DB, indices...). But, we want the option to skip re-installing JS dependencies.

- `invenio-cli update` should default to NOT installing the JS dependencies, because most of the time local files are changed rather than new dependencies installed. But we want the option to force the installation of new JS dependencies.

So we do want an install and don't install options in both commands but their default would be opposite... What are your thoughts on that?

setup.py
fenekku: Thanks for that! Changed.

invenio_cli/cli.py
fenekku: See discussion above. I am fine with revisiting once others try it out.

invenio_cli/cli.py
fenekku: Fixed (removed the `local` parameter). Yes, `invenio-cli services` would be enough and would just spin up the services needed in development. `invenio-cli containerize` would start them also if not running AND start the other production-like containers.

invenio_cli/cli.py
fenekku: Yes!

invenio_cli/cli.py
fenekku: I see... given that all commands (except for `demo` I believe) are only for 1 scenario, it doesn't make sense anymore to use the `Commands` abstraction. I made it consistent and the `LocalCommands` or `ContainerizedCommands` object is used in each case.

invenio_cli/helpers/commands.py
fenekku: I think I understand and I thought I had this covered, but you are right that if an error occurs in one of those forced commands, we get into a situation where `invenio-cli services` will keep raising an error: (concrete example just so we are on the same page)

Let's say the `invenio index destroy` underlying call raises an exception when running `invenio-cli services --force`, running `invenio-cli services --force` again will break, because now `invenio db destroy --yes-i-know` doesn't find the database! (it destroyed it last time around). `invenio-cli services` will break too for that matter since `invenio index init`  will complain we are trying to create a new one with the same name!

So wherever we put the set code, we still get a problem. Putting it at the end of the force block is a little better perhaps - moved it :+1:   

To solve our problems, we need an idempotent version of all those commands : 
- invenio db destroy: https://github.com/inveniosoftware/invenio-db/issues/126
- invenio index destroy: https://github.com/inveniosoftware/invenio-indexer/issues/114
- invenio index queue init purge : same as above
(and the regular ones too)

invenio_cli/helpers/commands.py
fenekku: Yes, once we add support for S3 at the installation level (cookiecutter), we can change that piece of code. :+1:

invenio_cli/helpers/cli_config.py
fenekku: > not a fan of asserts in the code

I hear you. If we were ever considering adding Python types to the code base, this would be a valuable place to put some.

... After thinking about it, we can remove it. On retrieval we use configparser's `getboolean` which will raise an Error if the value is not a boolean equivalent string. This way, even if someone passes a non-boolean (Python is dynamic, other things than boolean might be passed), on retrieval an error will be raised and the source of the problem will be identified.
Removing!

invenio_cli/helpers/commands.py
fenekku: Nope, not in `containerize`. Removing!

fenekku:
@ppanero and @zzacharo this is ready for review and for test drive. It is big so thanks for bearing with it.

README.rst
ppanero: > Marketing case

Agree :)

> demo command without option.

Yup, I hope. Is just we come back to the issue of where is `invenio-rdm-records` installed the venv or the container :( I was just wondering if you now see a way of solving this with the new CLI :)

(Feel free to shelf it)

invenio_cli/helpers/commands.py
ppanero: Fully agree, if we get them idempotent it should be fine with it at the end. Will push for those to get solved. Thanks!

README.rst
ppanero: I think it makes sense. In addition I do not see a problem with defaults being opposite (as long as is documented). My point being: a "default" is supposed to be that value that is used the most, in order to avoid typing  (we are lazy by nature)... So I think it all makes sense :)

README.rst
fenekku: Added it as a point on https://github.com/inveniosoftware/invenio-cli/issues/108

invenio_cli/cli.py
```diff
@@ -8,69 +8,15 @@
...
 import click
 
-from .helpers import DockerHelper, LoggingConfig, LogPipe, bootstrap, \
-    build_assets
-from .helpers import server as scripts_server
-from .helpers import setup as scripts_setup
-from .helpers import update_statics
 from .helpers.cli_config import CLIConfig
-from .helpers.commands import Commands, LocalCommands
+from .helpers.commands import Commands, ContainerizedCommands, LocalCommands
```
ppanero:
```suggestion
from .helpers.commands import ContainerizedCommands, LocalCommands
```

Unused import :)

invenio_cli/cli.py
```diff
@@ -107,68 +53,51 @@ def init(flavour):
...
-    click.secho('Setting up environment for {flavour} application...'
-                .format(flavour=invenio_cli.flavour), fg='green')
-
-    # Initialize docker client
-    docker_helper = DockerHelper(local=local,
-                                 log_config=invenio_cli.log_config)
+              help='Force recreation of db tables, ES indices, queues...')
+@click.option('--install-js/--no-install-js', default=True, is_flag=True,
+              help="(re-)Install JS dependencies, defaults to True")
+def containerize(force, install_js):
```
ppanero: Moderate: I could not make this run due to the lack of  `--pre` . I agree that it should not be necessary once we release. However, I imagine someone creating a package and wanting to test in containers after development.

In addition, sometimes releases are tedious and take time, so you want to have something in containers even if not fully released...

I would add the feature back, WDYT?

invenio_cli/helpers/commands.py
```diff
@@ -182,11 +192,18 @@ def services(self, force):
                 'pipenv', 'run', 'invenio', 'index', 'queue', 'init', 'purge']
             subprocess.run(command, check=True)
 
-        # If re-run, this doesn't throw an error
+            self.cli_config.update_services_setup(False)
+
+        if self.cli_config.get_services_setup():
```
ppanero: Minor: [Personal biased] I do not like bare returns.

What about:

``` python
if not self.cli_config.get_services_setup():
    [... All code inside this block ...]
```

invenio_cli/helpers/commands.py
```diff
@@ -195,22 +212,27 @@ def services(self, force):
         ]
         subprocess.run(command, check=True)
 
-        # If re-run, this throws an error
+        # Without the self.cli_config.get_services_setup() check
```
ppanero: Moderate: This comment is repeated for every command. I would just put it once on the top and explain it applies to all. Or is it like that so as they become idempotent we can take out one by one? If so I agree with it, but we should open a linked issue.

invenio_cli/helpers/commands.py
```diff
@@ -243,13 +265,122 @@ def signal_handler(sig, frame):
...
+
+        click.secho('Building assets...', fg='green')
+        self.docker_helper.execute_cli_command(
+            project_shortname, 'invenio webpack build')
+
+    def _lock_python_dependencies(self):
+        """Creates Pipfile.lock if not existing."""
+        locked = 'Pipfile.lock' in os.listdir('.')
+        if not locked:
+            subprocess.run(['pipenv', 'lock'], check=True)
```
ppanero: Moderate: Here is where I would add the `--pre` :) (with `if` clause :), not enforcing it always )

invenio_cli/helpers/commands.py
```diff
@@ -243,13 +265,122 @@ def signal_handler(sig, frame):
...
+
+    def containerize(self, force, install):
```
ppanero: Moderate: See above, add `--pre`?

invenio_cli/cli.py
fenekku: Thought so too, but it is still used for the `demo` command: https://github.com/inveniosoftware/invenio-cli/pull/106/files/0dc28a2ee9a3e4aec4a8f5d73b7c3e40214239a0#diff-95a0b99124ee2a80a5a1ae56da7acbf9R109

invenio_cli/cli.py
fenekku: Thanks for catching this: it wasn't on purpose! I had originally forgot to put locking back for `containerize` because running `invenio-cli install` locks things with `--pre` . Realizing `Pipfile.lock` would exist if `containerize` is run directly, I just added the locking back quickly without testing without prior install. I've added `pipenv lock --pre` back.

invenio_cli/helpers/commands.py
fenekku: Sure. Would be closer to what was did for `containerize` too.

invenio_cli/helpers/commands.py
fenekku: The latter was what I had in mind. I'll add the issue number for each, so it's clearer.

invenio_cli/helpers/commands.py
fenekku: Done :+1:

fenekku:
Updates made!

ppanero:
Tested `containerize`, works perfectly! Can be merged (already approved).

fenekku:
Needing to call docker-compose to stop the containers is not ideal. I've added an issue to add back stopping the containers after running `containerize`: https://github.com/inveniosoftware/invenio-cli/issues/109

PR #110, opened 2020-02-19
dependencies: update location creation command

ppanero:


ppanero:
The space at the end of the containerize fails sometimes and sometimes is needed. I'm investigating.

fenekku:
This PR should make those changes unnecessary: https://github.com/inveniosoftware/invenio-files-rest/pull/242 . (That being said maybe in the future we will only allow `invenio files location create` so we can put it in doesn't matter)

> space at the end of the containerize fails sometimes and sometimes is needed

This: https://github.com/inveniosoftware/invenio-cli/pull/110/files#diff-460932babb751f4442b83f1e96015e65R360 ? The space is definitely always needed; otherwise the two strings are concatenated and we get: 

"invenio files location create --default default-location${INVENIO_INSTANCE_PATH}/data"

invenio_cli/helpers/commands.py
```diff
@@ -357,7 +357,7 @@ def containerize(self, force, install):
             click.secho("Creating files location...", fg="green")
             self.docker_helper.execute_cli_command(
                 project_shortname,
-                "invenio files location --default default-location "
+                "invenio files location create --default default-location"
```
fenekku:
```suggestion
                "invenio files location create --default default-location "
```

setup.py
```diff
@@ -47,7 +47,7 @@
     'docker>=4.1.0,<4.2.0',
     'Flask-BabelEx>=0.9.4',
     'pipenv==2018.11.26',
-    'PyYAML>=5.1.2,<5.2.0',
+    'PyYAML>=5.1.2',
```
fenekku: I am thinking right now we should make packages less than the major version for third-party packages:

- follows semantic versioning, so if everybody plays by those rules, we are good
- if third-party packages don't play by those rules, the instance lock file should save us from their breaking changes

So what about something like:

`'PyYAML>=5.1.2,<6.0.0'` (or `'PyYAML~=5.1.2'`)

tests/test_helpers/test_commands.py
```diff
@@ -278,7 +278,7 @@ def test_containerizedcommands_containerize(
         call('project-shortname', 'invenio db init create'),
         call(
             'project-shortname',
-            'invenio files location --default default-location '
+            'invenio files location create --default default-location'
```
fenekku:
```suggestion
            'invenio files location create --default default-location '
```

invenio_cli/helpers/commands.py
ppanero: That's what I thought, I reverted back. But it was failing for me (before all the changes on files-rest so it's independent)...

Let's test more in depth after we decide what to do on S3.

tests/test_helpers/test_commands.py
ppanero: Same as before

setup.py
ppanero: From my perspective for external packages is better to leave them open. If we pin, we will be stuck forever in a version, because being realistic we do not have the manpower to be checking versions that come out of every library. And since the semantic versioning does not fully apply, we might miss a security patch  (what if a security patch is a breaking change?). From my perspective it's better to get at some point a breaking change rather than not get a security patch.

If something gets in place to follow us and use things that follow semantic versioning I'm happy to do so. But from my point of view it's and Invenio decision (like, lets have a talk with the architects) rather than here in a single module.

I know we did it in cookiecutter, but its for our modules during release since when we are stable we will depend on bundles.

fenekku: Let's videochat about it tomorrow or today.

invenio_cli/helpers/commands.py
```diff
@@ -382,3 +387,12 @@ def containerize(self, force, install):
 
         click.secho(
             'Instance running!\nVisit https://localhost', fg='green')
+
+    def demo(self):
+        """Add demo records into the instance."""
+        project_shortname = self.cli_config.get_project_shortname()
+
+        # The way of starting in container mode ensures the
+        # containers are up
```
fenekku: Minor: Probably safer and more consistent to have containers started again though...

invenio_cli/helpers/commands.py
```diff
@@ -205,8 +205,8 @@ def services(self, force):
             # this throws an error on re-runs
             # TODO: invenio-files-rest#238 should make it idempotent
```
fenekku: The comments can be removed now.

ppanero: fixed

invenio_cli/helpers/commands.py
ppanero: Shelved (as discusse IRL)

PR #111, opened 2020-02-19
Cookiecutter improvements: version, cookiecutter, folder error

### LINKED ISSUE #75: Provide better error message when outside directory
When outside the directory with the `.invenio` file. Most `invenio-cli` commands output:

```
ERROR:root:No flavour specified.
```

which is not as clear as `Missing .invenio file in current directory`. Use that error message instead.

### LINKED ISSUE #82: print invenio-cli version with invenio-cli --version
RIght now `invenio-cli --version` prints

```
Python 3.6.8
Flask 1.1.1
Werkzeug 0.16.0
```

It should print:

```
invenio-cli <version> 
```

### LINKED ISSUE #100: helpers [cookiecutter]: make cookiecutter version configurable

fenekku:
closes #100 
closes #75 
closes #82

### DIFF
--- a/invenio_cli/cli.py
+++ b/invenio_cli/cli.py
@@ -20,6 +20,7 @@ from .helpers.docker_helper import DockerHelper
 
 
 @click.group()
+@click.version_option()
 def cli():
     """Initialize CLI context."""
 
--- a/invenio_cli/cli.py
+++ b/invenio_cli/cli.py
@@ -28,12 +28,17 @@ def cli():
 @cli.command()
 @click.option('--flavour', type=click.Choice(['RDM'], case_sensitive=False),
               default='RDM', required=False)
-def init(flavour):
+@click.option('--template', required=False,
+              help='Cookiecutter path or git url to template')
+@click.option('--checkout', required=False,
+              help='Branch, tag or commit to checkout if --template is a git url')  # noqa
+def init(flavour, template, checkout):
     """Initializes the application according to the chosen flavour."""
     click.secho('Initializing {flavour} application...'.format(
         flavour=flavour), fg='green')
 
-    cookiecutter_wrapper = CookiecutterWrapper(flavour)
+    template_checkout = (template, checkout)
+    cookiecutter_wrapper = CookiecutterWrapper(flavour, template_checkout)
 
     try:
         click.secho("Running cookiecutter...", fg='green')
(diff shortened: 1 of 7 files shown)


invenio_cli/cli.py
```diff
@@ -20,19 +20,22 @@
...
 def cli():
     """Initialize CLI context."""
 
 
 @cli.command()
 @click.option('--flavour', type=click.Choice(['RDM'], case_sensitive=False),
               default='RDM', required=False)
-def init(flavour):
+@click.option('--cookiecutter', nargs=2, required=False,
+              help='Pass <cookiecutter git url> <branch> to select cookiecutter template')  # noqa
```
zzacharo: Is the url really needed? Is this intended to test cookiecutter fork urls?

invenio_cli/helpers/cookiecutter_wrapper.py
```diff
@@ -23,12 +23,34 @@
...
         self.template_name = None
         self.flavour = flavour
 
+        if self.flavour.upper() == 'RDM':
+            self.template = (
+                cookiecutter[0] or
+                'https://github.com/inveniosoftware/'
+                'cookiecutter-invenio-rdm.git'
+            )
+            self.template_name = self.extract_template_name(self.template)
```
zzacharo: Where is the `template_name` used/needed?

invenio_cli/cli.py
ppanero: For forks is a good idea. I see a potential problem on people testing random stuff, but this is meant for us and many times we have stuff on our fork...

A quick minor:

```suggestion
              help='Pass <cookiecutter git url> <branch/tag> to select cookiecutter template')  # noqa
```

invenio_cli/helpers/cookiecutter_wrapper.py
ppanero: It's used on the replay part. It does not appear here cuz there is no changes, if you open the full file is at the bottom :)

invenio_cli/cli.py
fenekku: Ah I see! Making it work for local cookiecutters would be even more useful. I will do that.

invenio_cli/helpers/cookiecutter_wrapper.py
fenekku: What he said :)

Issue #114, opened 2020-02-24
containerized always uses `--pre`


ppanero:
Due to the `lock` function

```python
def _lock_python_dependencies(self):
        """Creates Pipfile.lock if not existing."""
        locked = 'Pipfile.lock' in os.listdir('.')
        if not locked:
            subprocess.run(['pipenv', 'lock', '--pre'], check=True)
```

The containerized version always has dev/alphas in it. `--pre` should only be optional.

PR #116, opened 2020-02-25
containerize: make pre-release deps lock optional

ppanero:
Changes:
* Adds `pre` to containerize command to allow locking pre-releases only when specified.

Closes https://github.com/inveniosoftware/invenio-cli/issues/114


tests/test_helpers/test_commands.py
```diff
@@ -330,6 +330,17 @@ def test_containerizedcommands_containerize(
         expected_force_calls + expected_setup_calls
     )
 
+    # Case: pre=True, force=True, install=True
+    # Testing pre is only does not change the commands result.
```
fenekku:
```suggestion
    # Testing pre does not change the commands result.
```

Issue #121, opened 2020-03-18
init/install: "project_dir" shouldn't be an absolute path, but auto-detected


slint:
The fact that `project_dir` is filled-in during `init` as an absolute path, makes it not possible to re-use the generated project files in a different directory/machine. Example:

```bash
# In my machine (e.g. with "HOME=/home/slint")
$ invenio-cli init
...generate "zenodo-rdm" instance...
$ cd zenodo-rdm
$ cat zenodo-rdm/.invenio

[cli]
flavour = RDM
project_dir = /home/slint/src/zenodo-rdm
instance_path = /home/slint/Envs/zenodo-rdm-5ubStj9L/var/instance
logfile = /home/slint/src/zenodo-rdm/logs/invenio-cli.log
...
$ git init && git add . && git commit -m "initial commit" && git push ...

# in another machine (e.g. with "HOME=/home/glignos")
$ git clone https://gtihub.com/zenodo/zenodo-rdm.git && cd zenodo-rdm
$ invenio-cli install
$ invenio-cli install
Installing python dependencies...
...pipenv creates virutalenv correctly at /home/glignos/Envs/...
Symlinking invenio.cfg...
Symlinking templates/...
Collecting statics and assets...
...
Collect static from blueprints.
Created webpack project.
Installing js dependencies...
...
Installed webpack project.
Copying project statics and assets...
...fails with:

    distutils.errors.DistutilsFileError: cannot copy tree '/home/slint/src/zenodo-rdm/static': not a directory
```

A possible solution is to follow the behaviour of `pipenv` regarding the `Pipfile` location in order to auto-detect the `project_dir` value. Basically if you run `invenio-cli` in a directory that has a `.invenio` file, it should use this directory as the absolute path, instead of baking it in `.invenio`.

This probably also requires some design/refactoring on how the CLI determines its context...

ppanero:
Thanks for reporting! It makes sense (Added it to the board)

P.S. Happy to see zenodo-rdm ;)

fenekku:
@slint Oh my... thank you for that! Making sure the cli configuration/file is re-usable on different developer machine was a total blindspot on my part. Pablo and I have fixed a number of problems that was going to cause in https://github.com/inveniosoftware/invenio-cli/pull/125 .

PR #125, opened 2020-04-03
config: use cwd as project_dir

ppanero:
Closes https://github.com/inveniosoftware/invenio-cli/issues/121

Changes:

- `project_dir` is calculated based on where the cli is called from. This would have implications on how the logs file is set. It is now a relative path and should be added the `project_dir` as base. However, as far as I can see the logging config is not being created.
- Relaxes the dependency on click version. Invenio does not have it pin and it gave a pipenv collision with the new 7.1.1 released.

Tested:

- local: Ok
- containers: Ok

**Question**: Shall we follow the same approach with `instance_path` since it is caluclated for each case, is it worth having it written? I think we can just assume is the current env. Meaning, we keep calculating as till now, but we dont write it. It will be calculated on start up.

### DIFF
--- a/invenio_cli/helpers/cli_config.py
+++ b/invenio_cli/helpers/cli_config.py
@@ -36,6 +36,10 @@ class CLIConfig(object):
         try:
             with open(fullpath) as cfg_file:
                 self.config.read_file(cfg_file)
+
+            # Assign current dir a project dir. If reached here, it exists.
+            self.project_dir = Path(os.getcwd())
+
         except FileNotFoundError:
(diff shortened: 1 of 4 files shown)


fenekku:
I've made another commit fixing this blindspot. This problem was also going to affect the `services`' initial setup cache. I decided to have another file `.invenio.private` created that is not version controlled. This way we keep advantage of caching for some values. Goes with the added .gitignore PR https://github.com/inveniosoftware/cookiecutter-invenio-rdm/pull/65

PR #135, opened 2020-05-29
helpers: use 127.0.0.1 instead of localhost

lnielsen:
- Changes the output text for a successful running site from using
  https://localhost:5000 to https://127.0.0.1:5000, making it consistent
  with output from Flask. Also, Safari is not happy to go to localhost
  with SSL and changing certificates.


lnielsen:
Close in favour of #136

PR #136, opened 2020-06-01
cli: change --flavour to an argument

lnielsen:


README.rst
```diff
@@ -70,7 +70,7 @@ Containerized 'Production' environment
 .. code-block:: console
 
     # Initialize environment and cd into <created folder>
-    $ invenio-cli init --flavour=RDM
+    $ invenio-cli init rdm
```
zzacharo, line 73: Probably you need to use `RDM` here, right?

ppanero: It has `case_sensitive=False`, so I think `rdm` should work.

lnielsen: I personally prefer a lower-case version in the cli as it's one key less to press (shift). And yes, it's case insensitive.

tests/test_cli.py
```diff
@@ -0,0 +1,40 @@
...
+
+from invenio_cli.cli import cli
+
+
+@pytest.fixture()
+def runner():
+    """Click CLI runner."""
+    runner = CliRunner()
+    with runner.isolated_filesystem():
+        yield runner
```
fenekku, line 24: This is **very** nice. Thank you for adding that!

PR #144, opened 2020-07-22
helpers: improve exception message for project and machine configuration files

gqueiroz:


gqueiroz:
Looks like [Travis CI job](https://travis-ci.org/github/inveniosoftware/invenio-cli/jobs/710596603) failed in the first attempt due to the `isort` call during `./run-tests.sh` execution:
- `isort: error: argument -f/--future: expected one argument`

Maybe this is related to release of `isort 5`: [Upgrading to 5.0.0](https://timothycrosley.github.io/isort/docs/upgrade_guides/5.0.0/)

In the commit I suggested using this:
- `isort . --check-only --diff ` instead of `isort -rc -c -df` in `./run-tests.sh`.

but I am not sure if it would be safer to use a specific list of directories instead of `.` to avoid inclusion of *venvs*.

ppanero:
Hi @gqueiroz!

Thanks for the PR! IMHO I would use module name and tests, it's mainly where there are python files:

`isort invenio_cli tests --check-only --diff`

WDYT?

gqueiroz:
Hi @ppanero,

I totally agree!
May I change that in the PR and send it again?

ppanero:
Thanks a lot @gqueiroz! will be merged soon.

PR #148, opened 2020-08-12
Cli destroy command

### LINKED ISSUE #120: destroy command
Returns TODO: Revisit destroy command

cenouralm:
Closes #120

### DIFF
--- a/invenio_cli/cli.py
+++ b/invenio_cli/cli.py
@@ -146,7 +146,9 @@ def update(install_js):
               help='Verbose mode will show all logs in the console.')
 def destroy(local, verbose):
     """Removes all associated resources (containers, images, volumes)."""
-    click.secho('TODO: Revisit destroy command', fg='red')
+    cli_config = CLIConfig()
+    commands = Commands(cli_config, local)
+    commands.destroy()
 
 
 @cli.command()
--- a/invenio_cli/helpers/commands.py
+++ b/invenio_cli/helpers/commands.py
@@ -262,7 +262,6 @@ class LocalCommands(object):
         try:
             subprocess.run(['pipenv', '--rm'], check=True)
             click.secho('Virtual environment destroyed', fg='green')
-
         except subprocess.CalledProcessError:
             click.secho('The virtual environment was '
                         'not removed as it was not '
--- a/invenio_cli/helpers/commands.py
+++ b/invenio_cli/helpers/commands.py
@@ -257,6 +257,25 @@ class LocalCommands(object):
             'Instance running!\nVisit https://127.0.0.1:5000', fg='green')
         server.wait()
 
+    def destroy(self):
+        """Destroys a virtualenv (if created with pipenv) and containers."""
+        try:
+            subprocess.run(['pipenv', '--rm'], check=True)
+            click.secho('Virtual environment destroyed', fg='green')
+
+        except subprocess.CalledProcessError:
+            click.secho('The virtual environment was '
+                        'not removed as it was not '
+                        'created by pipenv', fg='red')
+
+        docker_helper = DockerHelper(
+            self.cli_config.get_project_shortname(),
+            local=True)
+
+        self.cli_config.update_services_setup(False)
+        docker_helper.destroy_containers()
+        click.secho('Destroyed containers...', fg='green')
+
 
 class ContainerizedCommands(object):
     """Containerized environment CLI commands."""
@@ -387,3 +406,17 @@ class ContainerizedCommands(object):
         # container mode ensures the containers are up
         self.docker_helper.execute_cli_command(
             project_shortname, 'invenio rdm-records demo')
+
+    def destroy(self):
+        """Destroys the instance's virtualenv and containers."""
+        try:
+            subprocess.run(['pipenv', '--rm'], check=True)
+            click.secho('Virtual environment destroyed', fg='green')
+        except subprocess.CalledProcessError:
+            click.secho('The virtual environment was '
+                        'not removed as it was not '
+                        'created by pipenv', fg='red')
+
+        self.docker_helper.destroy_containers()
+        self.cli_config.update_services_setup(False)
+        click.secho('Destroyed containers...', fg='green')
--- a/tests/test_helpers/test_commands.py
+++ b/tests/test_helpers/test_commands.py
@@ -16,6 +16,8 @@ import pytest
 from invenio_cli.helpers.commands import Commands, ContainerizedCommands, \
     LocalCommands
 
+from invenio_cli.helpers.docker_helper import DockerHelper
+
 
 def test_commands_delegates_to_environment():
     commands = Commands(Mock(), True)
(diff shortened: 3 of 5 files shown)


invenio_cli/cli.py
```diff
@@ -144,9 +144,11 @@ def update(install_js):
               help='Which environment to build, it defaults to local')
 @click.option('--verbose', default=False, is_flag=True, required=False,
               help='Verbose mode will show all logs in the console.')
-def destroy(local, verbose):
+def destroy(local, verbose, remove):
```
ppanero, line 147: There is no `remove` option above + this parameter is never used in the function.

```suggestion
def destroy(local, verbose):
```

invenio_cli/helpers/commands.py
```diff
@@ -151,8 +151,25 @@ def _ensure_containers_running(self):
         docker_helper.start_containers()
         # TODO: Find faster way to procede when containers are ready
         time.sleep(30)  # Give time to the containers to start properly
+    
+    def destroy(self):
+        """Destroys env, containers."""
+        command = [
+            'pipenv', '--rm'
+            ]
+        subprocess.run(command, check=True)
```
ppanero, line 160: This could be compressed in one line. Usually we divide it when is a large command, in this case is quite straight forward. It's mostly styling and me being OCD tho :)
```suggestion
        subprocess.run(['pipenv', '--rm'], check=True)
```

or if you want to be explicit about the command:

```suggestion
        command = ['pipenv', '--rm']
        subprocess.run(command, check=True)
```

invenio_cli/helpers/commands.py
```diff
@@ -151,8 +151,25 @@ def _ensure_containers_running(self):
...
+        click.secho('Destroyed Venv...', fg='green')
```
ppanero, line 161: Even though we use shortenings when talking among us, I guess for users is best to be explicit:
```suggestion
        click.secho('Virtual environment destroyed', fg='green')
```

invenio_cli/helpers/commands.py
```diff
@@ -151,8 +151,25 @@ def _ensure_containers_running(self):
...
+
+        docker_helper = DockerHelper(
+            self.cli_config.get_project_shortname(),
+            local=True)
+        """Stop containers"""
```
ppanero, line 166: I think both this and the below (destroy...) comments are not needed. Firstly I would say that function name is quite explicit. However, if you actually want to put the comment use `#`, `"""` is used for docstrings (top of the function) or multiline comments.
```suggestion
```

(Same for `"""Destroy containers"""`)

invenio_cli/helpers/commands.py
```diff
@@ -151,8 +151,25 @@ def _ensure_containers_running(self):
...
+        docker_helper.stop_containers()
+        """Destroy containers"""
+        docker_helper.destroy_containers()
+        click.secho('Destroyed containers...', fg='green')
 
-    def services(self, force):
+    def services(self, force, remove):
```
ppanero, line 172: Why is the argument needed? It is not used anywhere in the function + there is no option (click input) of it.

invenio_cli/cli.py
cenouralm: Forgot to remove! Thank you!

invenio_cli/helpers/commands.py
cenouralm: It's not needed, It was on a different implementation and I forgot to remove it. Nice catch!

invenio_cli/helpers/commands.py
```diff
@@ -151,6 +151,23 @@ def _ensure_containers_running(self):
         docker_helper.start_containers()
         # TODO: Find faster way to procede when containers are ready
         time.sleep(30)  # Give time to the containers to start properly
+    
+    def destroy(self):
+        """Destroys env, containers."""
```
ppanero, line 156: minor (self-OCD): Since there is space until the 80 characters let's be a bit more specific:
```suggestion
        """Destroys a virtualenv (if created with pipenv) and containers."""
```

Check if it fits in the 80 chars :)

invenio_cli/helpers/commands.py
```diff
@@ -151,6 +151,23 @@ def _ensure_containers_running(self):
...
+        try:
+            subprocess.run(['pipenv', '--rm'], check=True)
+            click.secho('Virtual environment destroyed', fg='green')
+
+        except:
+            pass
```
ppanero, line 162: major: We agreed IRL that this function call makes the CLI abort if the venv was not created by pipenv. So we need to catch it. However, it is not a good practice to simply `except` because you can be catching *any other* exception and the CLI will silently fail (this most likely ends-up in many hours of debugging and head aches). 

In addition, `pass` might sometimes be needed. But I would say *try to avoid it, because it looks fishy - you are hiding something*. In this case, I would suggest to use click to display a message to the user to inform of what happened: "The virtual environment was not removed as it was not created by pipenv". This way, first the user knows whats happening, and if something goes wrong it's easier for him to notify us :)

Therefore:
1- `except` just the type of exception needed
2- Display a message in the except.

cenouralm: This https://github.com/inveniosoftware/invenio-cli/pull/148/commits/3e9dec8500cbd8022a148f978db66ae1465139b4  solves it :)

invenio_cli/helpers/commands.py
```diff
@@ -333,6 +333,25 @@ def _lock_python_dependencies(self, pre):
         if not locked:
             subprocess.run(command, check=True)
 
+    def destroy(self):
+        """Destroys a virtualenv (if created with pipenv) and containers."""
```
ppanero, line 337: major: Try to *never repeat code*. If something is repeated, it means *it has to be extracted into a function*. My point being, this and `LocalCommands.destroy` is 100% the same, should be written only once.

invenio_cli/helpers/commands.py
```diff
@@ -333,6 +333,25 @@ def _lock_python_dependencies(self, pre):
...
+            click.secho('Virtual environment destroyed', fg='green')
+
+        except subprocess.CalledProcessError:
+            click.secho('The virtual environment was '
+                        'not removed as it was not '
+                        'created by pipenv', fg='red')
+
+        docker_helper = DockerHelper(
+            self.cli_config.get_project_shortname(),
+            local=True)
```
ppanero, line 349: 🤔 `local=True` in containerize... suspicious ;)

When you refactor this into a function, you might be tempted to think *oh this is different*. But not really... Could it just be a parameter forwarding solution? ;)

invenio_cli/cli.py
```diff
@@ -147,7 +147,12 @@ def update(install_js):
 def destroy(local, verbose):
     """Removes all associated resources (containers, images, volumes)."""
     cli_config = CLIConfig()
-    commands = LocalCommands(cli_config)
+
+    if local:
+        commands = LocalCommands(cli_config)
+    else:
+        commands = Commands(cli_config, local)
```
ppanero, line 154: `Commands` is a factory of "sub-commands" classes (i.e. `LocalCommands` and `ContainerizeCommands`. The *if-else* might not be needed ;).

Check how the `Commands` constructor works, and how the `demo` command is implemented. This, is a good first step :), but it can be much simpler by properly using the factory.

invenio_cli/helpers/commands.py
```diff
@@ -43,6 +43,25 @@ def __getattr__(self, name):
         """Delegate commands according to environment."""
         return getattr(self.environment, name)
 
+    def destroy(self, local):
```
jbenito3, line 46: I wonder if there should be a default value for the parameter `local`?

cenouralm: In my understanding, the `local` parameter changes according to the environment. What would be the reason for having it defaulted?

jbenito3: it says in the definition, `--local` defaults to 'local', where is the defaulting happening? sorry...i didnt check in detail. Also, before you had 2 identical functions, one using `local` and one didn't. That triggered my question that maybe local is optional...

cenouralm: No worries! Still thank you for looking into my code :)

tests/test_helpers/test_commands.py
```diff
@@ -258,6 +258,23 @@ def test_localcommands_demo(
...
+def test_localcommands_destroy(
+        docker_helper_subprocess, patched_docker_helper, patched_subprocess,
+        fake_cli_config):
+    commands = LocalCommands(fake_cli_config)
+
+    commands.destroy()
+
+    patched_subprocess.run.assert_called_with(['pipenv', '--rm'], check=True)
+
+    docker_helper_subprocess.run.destroy_containers()
```
fenekku, line 273: With the changes above we can actually test now:

```suggestion
    docker_helper_subprocess.run.assert_called_with(['docker-compose', '--file', 'docker-compose.yml', 'down', '--volumes'])
```

tests/test_helpers/test_commands.py
```diff
@@ -258,6 +258,23 @@ def test_localcommands_demo(
     )
 
 
+@patch('invenio_cli.helpers.commands.subprocess')
+@patch('invenio_cli.helpers.commands.DockerHelper')
```
fenekku, line 262: remove this

invenio_cli/helpers/commands.py
fenekku: We ended up putting "very similar" code in LocalCommands and ContainerizedCommands to get things moving. We can refactor this later when the higher-level commonality is even clearer.

fenekku (approved):

PR #153, opened 2020-09-01
Cenouralm destroycommand

fenekku:
@cenouralm Check it out and use it to implement the localized version: bring these changes in your branch and add the localized version.


fenekku:
was for demonstration purposes.

PR #157, opened 2020-09-16
Introduce command to stop local containers

cenouralm:


invenio_cli/cli.py
```diff
@@ -122,14 +122,16 @@ def demo(local):
 
 
 @cli.command()
-def run():
+@click.option('--stop', default=False, is_flag=True,
+              help='Stop containers')
+def run(stop):
```
ppanero, line 127: After discussing IRL, the stop should be a separate *command*, not a flag on the *run*. (We we called option b when we talked)

```suggestion
def run():
```

invenio_cli/helpers/commands.py
```diff
@@ -228,8 +228,16 @@ def demo(self):
         command = ['pipenv', 'run', 'invenio', 'rdm-records', 'demo']
         subprocess.run(command, check=True)
 
-    def run(self):
+    def run(self, stop):
         """Run development server and celery queue."""
+        if stop:
+            command = [
```
ppanero, line 234: There is a `stop_containers` function in the `docker_helper` file. It might be good using that :)

cenouralm: Will do!

invenio_cli/cli.py
```diff
@@ -122,6 +122,8 @@ def demo(local):
 
 
 @cli.command()
+@click.option('--stop', default=False, is_flag=True,
+              help='Stop containers')
```
ppanero, line 126: Not needed
```suggestion
```

invenio_cli/cli.py
```diff
@@ -154,6 +156,18 @@ def destroy(local, verbose):
...
 
+@cli.command()
+@click.option('--local/--containers', default=True, is_flag=True,
+              help='Which environment to build, it defaults to local')
+@click.option('--verbose', default=False, is_flag=True, required=False,
+              help='Verbose mode will show all logs in the console.')
+def stop(local, verbose):
+    """Stops containers."""
+    cli_config = CLIConfig()
+    commands = Commands(cli_config, local)
```
ppanero, line 167: Nice use of oop 🚀

tests/test_helpers/test_commands.py
```diff
@@ -295,7 +295,7 @@ def test_localcommands_run(
         fake_cli_config):
     commands = LocalCommands(fake_cli_config)
 
-    commands.run()
+    commands.run(stop=False)
```
ppanero, line 298: Not re-reviewing tests. Not updated as mentioned IRL.

invenio_cli/helpers/commands.py
```diff
@@ -334,11 +343,6 @@ def _lock_python_dependencies(self, pre):
 
     def containerize(self, pre, force, install, stop):
```
ppanero, line 344: Remove also the parameter since it's not used. No?
```suggestion
    def containerize(self, pre, force, install):
```

invenio_cli/helpers/commands.py
```diff
@@ -228,6 +228,15 @@ def demo(self):
         command = ['pipenv', 'run', 'invenio', 'rdm-records', 'demo']
         subprocess.run(command, check=True)
 
+    def stop(self):
+        """Stops containers."""
+        docker_helper = DockerHelper(
```
ppanero, line 233: LGTM. This is more a comment for future work:

We should try to find a common way of using `docker_helper`. The containerize command has an instance as attribute and the local commands instantiate it each time is needed. It *might* be able to be refactored into something common.

invenio_cli/helpers/commands.py
```diff
@@ -228,6 +228,15 @@ def demo(self):
...
+            self.cli_config.get_project_shortname(),
+            local=True)
+
+        docker_helper.stop_containers()
+        click.secho('Stopped containers...', fg='green')
```
ppanero, line 238: [Full OCD]: IMO the ... are sort of an indication of something going on - in progress -. Therefore, I would use if for the messages *before* the action. e.g. "creating indexes...". 

For messages of things that *are already finished* I would not put the ... For example https://github.com/inveniosoftware/invenio-cli/blob/28a2e17ac9cba094ba018b3fb1e25105edfcaedf/invenio_cli/helpers/commands.py#L427

If you decide to change this (totally up to yo, as in, this is my OCD). Could you change the `Destroyed ...` messages too and remove the ...

```suggestion
        click.secho('Containers stopped', fg='green')
```

tests/test_helpers/test_commands.py
```diff
@@ -312,6 +312,28 @@ def test_localcommands_run(
     assert patched_subprocess.Popen.mock_calls == expected_calls
 
 
+@patch('invenio_cli.helpers.commands.subprocess')
+@patch('invenio_cli.helpers.docker_helper.subprocess')
+def test_localcommands_stop(
+        patched_docker_helper, patched_subprocess, fake_cli_config):
+    commands = LocalCommands(fake_cli_config)
+
+    # Needed to fake call to docker-compose --version but not call
```
ppanero, line 321: Doubt: What does this comment applies to?

tests/test_helpers/test_commands.py
```diff
@@ -312,6 +312,28 @@ def test_localcommands_run(
     assert patched_subprocess.Popen.mock_calls == expected_calls
 
 
+@patch('invenio_cli.helpers.commands.subprocess')
+@patch('invenio_cli.helpers.docker_helper.subprocess')
+def test_localcommands_stop(
```
ppanero, line 317: moderate: There is no assertion in this test, in case something does not work at desired what kind of feedback do we get when running tests? How is it notified? stacktrace?

tests/test_helpers/test_commands.py
cenouralm: This comment was made when developing the `test_localcommands_destroy` and I can't really remember the reason why was done.

tests/test_helpers/test_commands.py
cenouralm: The new commit addressed the missing assertion https://github.com/inveniosoftware/invenio-cli/pull/157/commits/63801446b68c62d7a896981a2a7b17eb22d41afd

invenio_cli/helpers/docker_helper.py
```diff
@@ -57,11 +57,12 @@ def start_containers(self):
...
-        if self.local:
-            command[2] = 'docker-compose.yml'
+        command = [
+            'docker-compose',
+            '--file',
+            'docker-compose.yml' if self.local else 'docker-compose.full.yml',
+            'stop'
+        ]
 
         subprocess.call(command)
```
fenekku, line 67: Turns out the same trick @lnielsen mentioned can be used for the stop command too (It works. I learned something!): 

```python
def stop_containers(self):
    command = [
        'docker-compose',
        '--file',
        'docker-compose.full.yml',
        'stop'
    ]
    subprocess.call(command)
```


Testing becomes much easier and uniform too.

fenekku (commented):
Applying Lars's trick is the way to go - will cause some refactoring of this PR but it's for the best.

.travis.yml
```diff
@@ -24,11 +28,13 @@ cache:
 
 env:
   - REQUIREMENTS=lowest
-  - REQUIREMENTS=release DEPLOY=true
+  - REQUIREMENTS=release
   - REQUIREMENTS=devel
 
 python:
-  - "3.6"
+  # Lower than 3.6.9 distutils imports `imp` and throws a deprecation
```
ppanero, line 35: Is this the case for this library?

cenouralm:
The successor of this PR is #157

Issue #160, opened 2020-10-01
celery worker not starting


ppanero:
Package version (if known): 0.17.1

## Steps to Reproduce
1. run `invenio-cli run`
2. See error in the console:
```
Usage: celery worker [OPTIONS]
Try 'celery worker --help' for help.

Error: no such option: --app
```

## Expected behavior

Celery worker to be up and running

jhrdt:
Hej,

I had the same issue lately.
If you use the latest celery version (5.0.0), the order of the cli arguments changed.

See PR #161

ppanero:
@wgresshoff reported this happening again in v0.17.2

ppanero:
delegated to helm chart, local and containerize are working

PR #161, opened 2020-10-04
Fix worker invokation for celery >= 5.0.0

jhrdt:
Celery 5.0.0 changed the order of cli arguments.

`--app` is no part of the worker subcommand, it belongs to the parent command.

A invocation of would look like `celery --app=proj worker ...`

See #160

### DIFF
--- a/invenio_cli/helpers/commands.py
+++ b/invenio_cli/helpers/commands.py
@@ -242,7 +242,7 @@ class LocalCommands(object):
 
         click.secho("Starting celery worker...", fg="green")
         worker = subprocess.Popen([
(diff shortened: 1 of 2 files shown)


ppanero (approved):

lnielsen:
Will this work with Celery 4.4? I'm asking because Celery has named 4.4 as a Long-Term Support release, and said that in new 5.x releases there'll be lots of breaking changes, thus preferably we should work with both 4.x and 5.x

jhrdt:
I works with Celery >= 5 .

lnielsen:
I've just checked, and it seems like it should work on celery 4.4 as well.

PR #168, opened 2020-10-27
commands: add site hostname config when running local

ppanero:
**Open question**
- Do we assume that there is always a port and host? e.g. if I want to run in `mysite.com` I also provide the port (no 443 assumptions if `port == None`)

### DIFF
--- a/invenio_cli/helpers/commands.py
+++ b/invenio_cli/helpers/commands.py
@@ -248,6 +248,7 @@ class LocalCommands(object):
         click.secho("Starting up local (development) server...", fg='green')
         run_env = os.environ.copy()
         run_env['FLASK_ENV'] = 'development'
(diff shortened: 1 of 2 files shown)


fenekku (approved):
As for the question about ports:

This is for development mode, so we should clearly dissuade developers from using that mode in actual production. (`run_env['FLASK_ENV'] = 'development'` is a no-no).

ppanero:
@fenekku I see my example was not good 😅 I didn't mean production, that would be wreckage before departing hehe. My point was that users might want to use maybe 80 or 443 in their machines (in a similar fashion than containerize). But okay, I see this not happening since for that it needs a Nginx/similar in front so it can keep the 5000.

ppanero:
closes https://github.com/inveniosoftware/cookiecutter-invenio-rdm/issues/111

Issue #170, opened 2020-10-30
global: simplify development 


lnielsen:
Following is a quick brainstorming on commands that potentially could be useful for making it easier to develop with RDM:

Boostrap and install RDM locally:
```
invenio-cli init rdm
cd ~/src/my-site
invenio-cli install --pre
```

Install a Python and JS module for development:
```
invenio-cli module-install -e ~/src/invenio-records-resources
invenio-cli js-install -e ~/src/react-invenio-forms
```

Run:
```
invenio-cli services
invenio-cli run
```

Watch JS changes if needed:
```
invenio-cli watch ~/src/react-invenio-forms
```

Alternative assets commands:
```
invenio-cli assets clean
invenio-cli assets buildall
```

fenekku:
I was thinking about the extensions (https://inveniordm.docs.cern.ch/extensions/) and how the install commands actually fall in that category of functionality.

So having 

`invenio-cli ext init`  To create your own.
`invenio-cli ext js-install ...`  To install
`invenio-cli ext module-install ...` 

could be an idea..

lnielsen:
What should ``invenio-cli ext module-install <PATH>`` do?

```
pipenv run pip install -e <PATH>
```

Should it also run update?
Do we need to write something to the ``.invenio`` file?

lnielsen:
In #173 I've added the commands:

```
invenio assets update --force --development
invenio assets watch
```

The ``--development`` or ``--production`` will respectively set FLASK_ENV=development/production to ensure we either copy or symlinkfiles.
The ``--force`` replaces ``--instal-js/--no-install-js``. Force will clean the folder and install node modules. If not, only a invenio webpack create/build is run (this is needed to wipe assets if e.g. it was created with production and you now want development mode).


For the React module development I'm thinking instead of ``invenio-cli ext js-install ~/src/react-invenio-deposit``, we could have something like:

```
invenio-cli assets watch-module ~/src/react-invenio-deposit
```

This would:

- Install and build module: ``cd ~/src/react-invenio-deposit; npm install; npm run build``
- Link module ``cd ~/src/react-invenio-deposit; npm run link-dist``
- Link project ``cd <path-to-your-virtualenv>/var/instance/assets/node_modules; npm link react-invenio-deposit``
- Start watching changes: ``cd ~/src/react-invenio-deposit; npm run watch``

You will still need afterwards to run:

```
invenio-cli assets watch
```

lnielsen:
[...]

PR #173, opened 2020-11-04
commands: assets update/watch/install-module/watch-module & shell & pyshell 

lnielsen:
* Moves ``invenio-cli update`` to ``invenio-cli assets update``

* Adds options ``invenio-cli assets update --force --development``: 
   - ``--force``: Will clean the assets folder including node modules and install/build the webpack project.
   - ``--development``: Will symlink files (by setting ``FLASK_ENV=development``)

* Adds command ``invenio-cli assets watch`` to watch the instance's webpack project for changes (running ``pipenv run invenio webpack  run start``).

* Adds command ``invenio-cli assets install-module`` which will install, build, and link a JavaScript module (using pynpm module).

* Adds command ``invenio-cli assets watch-module`` which watch changes on a JavaScript module.

* Adds command ``invenio-cli shell`` which will start a virtualenv shell.

* Adds command ``invenio-cli pyshell`` which will start a Python application shell.

* Adds option ``invenio-cli install --pre --development`` to set FLASK_ENV=developement and thereby symlink assets

Addresses #170 

---

WIth this you can now:

```
invenio-cli init rdm -c master
invenio-cli install --pre
# Next line to be replaced with "invenio-cli ext module-install ~/src/invenio-rdm-records"
pipenv run pip install -e ~/src/invenio-rdm-records -e ~/src/invenio-app-rdm 
invenio-cli assets update --force --development
invenio-cli services
invenio-cli demo

# New shell
invenio-cli run

# New shell
invenio-cli assets watch-module ~/src/react-invenio-deposit --build

# New shell
invenio-cli assets watch
```

Now, live happily ever after with live reloading when editing files.

---

### Note 1

I didn't go with ``invenio-cli ext js-install`` as a Python module install and React module install works very differently, so I personally felt it was more appropriate to have an ``assets`` command which operates purely on the assets.

### Note 2

We could likely integrate ``invenio-cli assets update --force --development`` into ``invenio-cli ext module-install ~/src/invenio-rdm-records``. For instance:

```
invenio-cli ext module-install ~/src/invenio-rdm-records ~/src/invenio-app-records
```

Would run:

```
pipenv run pip install -e ~/src/invenio-rdm-records -e ~/src/invenio-app-rdm 
invenio-cli assets update --force --development
```

You could disable the rebuild of assets with :

```
invenio-cli ext module-install --without-assets ~/src/invenio-rdm-records
```


ppanero:
To be crosschecked but if the `FLASK_ENV=development` is not on the install step it will actually do a hard copy of the assets then the second time (when `update`) it won't do anything (probably the `clean_create_cmd` is fixing this, but it requires always `--force`)

The first time is called here: https://github.com/inveniosoftware/invenio-cli/blob/6068cbac8fd6e0ab7de9d14034511910c0a34641/invenio_cli/helpers/commands.py#L164 and the default for the `flask_env` is `production`. Since it it a function inside `LocalCommands`, would it make sense to change default to `developement`? I would assume that whoever goes through installing locally would develop...

Just a heads up in case someone tries it, I think it is:
```
invenio-cli assets update --force --development
invenio-cli assets watch
```

Not `invenio`.

lnielsen:
> To be crosschecked but if the FLASK_ENV=development is not on the install step it will actually do a hard copy of the assets [...]

So, I've added (setting ``FLASK_ENV=development`` during install):

```
invenio-cli install --development
```

However, as soon as you install an existing module in dev mode (e.g. ``pipenv run pip install -e ~/src/invenio-app-rdm``), you need to force rebuild all assets (because the ``invenio webpack create`` is not smart enough to wipe an existing wrong link).

I think it makes sense to stay with ``production`` as default, since most users will just need to edit the assets in their own instance (which are always symlinked).

> Just a heads up in case someone tries it, I think it is:

Correct, I've edited the description accordingly.

invenio_cli/helpers/commands.py
```diff
@@ -110,37 +113,105 @@ def _copy_statics_and_assets(self):
...
+        watch_cmd = prefix + ['invenio', 'webpack', 'run', 'start']
+
+        with env(FLASK_ENV='development'):
+            # Collect into statics/ and assets/ folder
+            click.secho('Starting assets watching (press CTRL+C to stop)...',
+                        fg='green')
+            subprocess.run(watch_cmd, check=True)
+
+    @staticmethod
+    def _build_js_module(module_pkg):
```
lnielsen, line 160: Minor: Comment from @zzacharo that this is not needed.

PR #174, opened 2020-11-05
cli: add command to install modules

cenouralm:


invenio_cli/helpers/commands.py
```diff
@@ -64,6 +64,13 @@ def _install_py_dependencies(self, pre, lock):
             command += ['--skip-lock']
         subprocess.run(command, check=True)
 
+    def install_modules(self, modules):
+        """Install modules."""
+        single_module = 0
+        for single_module in modules:
+            subprocess.run(['pipenv', 'run', 'pip',
```
lnielsen, line 71: for efficiency we should run it as a single command:

pipenv run pip install -e mod1 -e mod2 -e mod3

instead of 

pipenv run pip install -e mod1 
pipenv run pip install -e mod2 
pipenv run pip install -e mod3

lnielsen: This should do the job:
```
subprocess.run(['pipenv', 'run', 'pip', 'install'] + [f'-e {m}' for m in modules], check=True")
```

invenio_cli/helpers/commands.py
```diff
@@ -64,6 +64,13 @@ def _install_py_dependencies(self, pre, lock):
             command += ['--skip-lock']
         subprocess.run(command, check=True)
 
+    def install_modules(self, modules):
+        """Install modules."""
+        single_module = 0
```
lnielsen, line 69:
```python
if len(modules) < 1:
    raise click.UsageError("You must specify at least one module.")
```

Issue #180, opened 2020-11-20
global: reorganize commands


ppanero:
# Invenio-CLI 

## Local

**First initial use case**: I am a new developer and want to set up my instance.

```
invenio-cli check-requirements
invenio-cli init rdm
cd my-site
invenio-cli packages lock
invenio-cli install
invenio-cli services setup # Will start services
invenio-cli run
invenio-cli services stop
```

Note: `invenio-cli install` will allow `--pre` and `--lock`, this will avoid the `invenio-cli packages lock` command. However, the install command will default to `--skip-lock`.

**Continuous development**: I want to re-take my work on a development instance.

```
invenio-cli services start
invenio-cli run
invenio-cli services stop
```

Note: Without the `run` it would work to but up services to e.g. use a pyshell with connected services.

**No demo content**: I want to install a dev environment without content since I will use my own records.

```
invenio-cli services start
invenio-cli services setup --force --no-demo-data  # Will not start services since we used start before (for presentation sake)
invenio-cli services stop
invenio-cli services status
invenio-cli services destroy # Only services: compose down
```

**External services**: I want to but a dev instance connected to external services

```
invenio-cli run --no-services # Dan's use case
```

**Other commands**

They stay as they are

```
invenio-cli assets update
invenio-cli assets watch
invenio-cli destroy 
```

They are moved

```
invenio-cli ext ---> invenio packages 
```

## Containerize

**I want to preview**: Minimal set of commands
Q: What about invenio-cli preview
```
invenio-cli init rdm
cd my-site
invenio-cli containers start --build --setup --lock
```

**I want to preview**: Step by step splitted

Q: What about not having flags, commands are locked to specific behaviours, no include through flags
```
invenio-cli init rdm
cd my-site
invenio-cli packages lock
invenio-cli containers build # Build application image including assets
invenio-cli containers setup --force [--no-demo-data]
invenio-cli containers start --no-services # (same name as run)
invenio-cli containers stop
invenio-cli containers destroy # Only services: compose down
```

Note: When building the image, assets will always be built. This will result in the removal of the `$include_assets` flag on the current Dockerfiles.
What does this mean?
- No explicit `update` command. It was tested and if you want to re-build assets, just re-building the docker image works the same since the pythong deps locking and installation is cached.
- Use cases where a change is needed but no re-build is required are quite limited to `invenio.cfg` changes and it was agreed to leave this as a future issue.

[...]

PR #182, opened 2020-11-24
global: command-steps pattern and cli groups

ppanero:
**Requires** https://github.com/inveniosoftware/cookiecutter-invenio-rdm/pull/124

**Changes**

- Refactor commands into its own module and split by type
- Refactor cli and split by group to improve readability
- Abstracted process calling and response types (e.g. `ExecResult` from docker and compose, others from `subprocess`, etc.) now are encapsulated in a `ProcessResponse`. Same is done for calls.
- Most of the commands, now return a series of steps. Except some like (`ensure_containers_are_running`, `update_statics_and_assets` and others that turned more complex when a step requires the output of the previous or use complex env vars.). Then the CLI `click.commands` are running this step and showing feedback to the user.

*Tests related*
- Split tests per case instead of having several cases in one test
- Minimize imports (this greatly helped to know what and why something was being patched).
- Caveat: Had to skip some tests, for now, tested flows manually and they work.

**Triggers**
- When agreed to merge this, `FIXME`s need to be turned into issues (e.g. turn the above complex operations into steps)

**Reference Documentation**
https://github.com/inveniosoftware/docs-invenio-rdm/pull/85

Closes https://github.com/inveniosoftware/invenio-cli/issues/180


invenio_cli/cli/assets.py
```diff
@@ -0,0 +1,87 @@
...
+        force=force,
+        flask_env='production' if production else 'development'
+    )
+
+
+@assets.command()
+@click.argument('path', type=click.Path(exists=True))
+@pass_cli_config
+def install(cli_config, path):
+    """Install and link a React module."""
```
lnielsen, line 43: Comment: Not sure, but perhaps this should mention it's fro development?

ppanero: Done, specified for all assets and watch commands.

invenio_cli/cli/assets.py
```diff
@@ -0,0 +1,96 @@
...
+          `invenio-cli containers build`.
+    """
+
+
+@assets.command()
+@click.option('--force', '-f', default=False, is_flag=True,
+              help='Force the full recreation the assets and statics.')
+@click.option(
+    '--production/--development', '-p/-d', default=True, is_flag=True,
+    help='Production mode copies files. Development mode symlinks files.'
```
fenekku, line 32: Let's indicate in the help message that `--production` is the default option.

dfdan:
The --no-services flag is a great addition.

Would it be possible to add (commented out) external service config options to invenio.cfg to help people use this?

PR #186, opened 2020-11-26
install: updates instance path before symlinking

Glignos:


Glignos:
Wraps the instance path calculation to happen after the python package installing.
Otherwise, the error for the missing bin was masked and the `result.output`from the pipenv command was an empty string. 
Leading for the working directory to be pointed as the instance one leading to circular pointers.

lnielsen:
Tested and working 👍

PR #201, opened 2020-11-30
global: fix wording/typos/1 clashing option

fenekku:


fenekku:
I think the Github-Action team ran into something similar with the CI @diegodelemos @lnielsen

invenio_cli/commands/services.py
```diff
@@ -34,21 +34,21 @@ def ensure_containers_running(self):
             project_shortname=project_shortname,
         )
         return ProcessResponse(
-                output="Containers start and checkup run.",
-                status_code=0,
+            output="Containers started and healthy.",
```
ppanero, line 37: comment: this is not 100% true since ES will still show a warning due to not being in cluster mode (status=yellow).

invenio_cli/cli/services.py
```diff
@@ -36,7 +36,7 @@ def start(cli_config):
 @services.command()
 @click.option('-f', '--force', default=False, is_flag=True,
               help='Force recreation of db tables, ES indices, queues...')
-@click.option('-n', '--no-demo-data', default=False, is_flag=True,
+@click.option('-N', '--no-demo-data', default=False, is_flag=True,
```
ppanero, line 39: Nice catch! could you also change it in the [containers](https://github.com/inveniosoftware/invenio-cli/blob/9227f0607aab925a7d02a7134b54a02314409af1/invenio_cli/cli/containers.py#L44) commands? cuz it happens too. Thanks!

ppanero (approved):

invenio_cli/cli/services.py
fenekku: :100:

invenio_cli/commands/services.py
fenekku: close enough :smile_cat:

fenekku:
Updates made. I am merging despite CI (I assume the issue will be fixed in the future).

Issue #199, opened 2020-11-30
Improve options experience


fenekku:
We can improve the command options.

### Consistency
Example:

invenio-cli packages lock --dev

but 

invenio-cli check-requirements --development

Let's use --dev or --development throughout. Sometimes those have slightly different meanings: either we choose better names or keep with consistent names.

### Defaults in help text
Example

```console
invenio-cli run --help
Usage: invenio-cli run [OPTIONS]

  Starts the local development server.

  NOTE: this only makes sense locally so no --local option

Options:
  -h, --host TEXT                 The interface to bind to.
  -p, --port INTEGER              The port to bind to.
  -d, --debug / --no-debug        Enable/disable debug mode including auto-
                                  reloading (default: enabled).

  -s, --services / -n, --no-services
                                  Enable/disable dockerized services (default:
                                  enabled).

  --help                          Show this message and exit.
```

Would be nice to have default host and port value listed.

Another example:

```console
invenio-cli install --help
Usage: invenio-cli install [OPTIONS]

  Installs the  project locally.

  Installs dependencies, creates instance directory, links invenio.cfg +
  templates, copies images and other statics and finally builds front-end
  assets.

Options:
  --pre                           If specified, allows the installation of
                                  alpha releases

  -p, --production / -d, --development
                                  Production mode copies statics/assets.
                                  Development mode symlinks statics/assets.

  --help                          Show this message and exit.
```

Which of `--production` or `--development` is default? (`--development` should be default)

Also we can try to use a consistent "(default: <bla>)" format used in many places. Although we sometimes use `(default: True)` and sometimes (default: enabled)` and sometimes `(default=True)`, so we can make those consistent too.

### Unique options

Some options are clashing:

```console
invenio-cli services setup --help
Usage: invenio-cli services setup [OPTIONS]

  Setup local services.

Options:
  -f, --force                     Force recreation of db tables, ES indices,
                                  queues...

  -n, --no-demo-data              Disable the creation of demo data
  --stop-services                 Stop containers after setup.
  -s, --services / -n, --no-services
                                  Enable/disable dockerized services (default:
                                  enabled).

  --help                          Show this message and exit.
```

What does `invenio-cli services setup -n` mean? Answer: --no-services . There might be other occurrences.

PR #212, opened 2021-01-15
install: allow locking dependencies on first installation

### LINKED ISSUE #198: Would like not to have to run invenio-cli packages lock before install
Package version (if known): november release

Documentation was missing about this (fixed in submitted PR): installing person needs to run `invenio-cli packages lock` before being able to run `invenio-cli install`. Ideally, `invenio-cli install` would just take care of things (re-introducing `--skip-lock` if want to do other install steps but not re-fetch-lock could be possible). 

Actually ran into this while upgrading: I changed my dependencies in my Pipfile and ran `invenio-cli install` like last time, but my changes to Pipfile were not taken into account. Not the ideal experience. My take is that we should limit the number and verbosity of commands installing people have to run.  

(It's tricky to understand the intention with the skipped tests...)

ppanero:
closes #198

### DIFF
--- a/invenio_cli/commands/install.py
+++ b/invenio_cli/commands/install.py
@@ -21,14 +21,14 @@ class InstallCommands(LocalCommands):
         """Constructor."""
         super(InstallCommands, self).__init__(cli_config)
 
-    def install_py_dependencies(self, pre):
+    def install_py_dependencies(self, pre, dev=False):
         """Install Python dependencies."""
-        steps = [
-            FunctionStep(
-                func=PackagesCommands.is_locked,
-                message="Checking if dependencies are locked."
-            )
-        ]
+        # If not locked, lock. Then install.
+        steps = []
+
+        if PackagesCommands.is_locked().status_code > 0:
+            steps.extend(PackagesCommands.lock(pre, dev))
+
         steps.extend(PackagesCommands.install_locked_dependencies(pre))
 
         return steps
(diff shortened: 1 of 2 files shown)


invenio_cli/commands/install.py
```diff
@@ -21,14 +21,14 @@ def __init__(self, cli_config):
...
-                message="Checking if dependencies are locked."
-            )
-        ]
+        # If not locked, lock. Then install.
+        steps = []
+
+        if PackagesCommands.is_locked().status_code > 0:
+            steps.extend(PackagesCommands.lock(pre, dev))
+
         steps.extend(PackagesCommands.install_locked_dependencies(pre))
```
fenekku, line 32: This command should also take `dev` then if we introduce `dev` support, right?

ppanero: Changes added. Note that there is one `dev` and `developement` flags... I agree that is an issue that should be addressed, but will be done as part of: https://github.com/inveniosoftware/invenio-cli/issues/199

lnielsen (approved):

invenio_cli/commands/install.py
fenekku: Ah, my comment was for `PackagesCommands.install_locked_dependencies(pre)` i.e. it should be: `PackagesCommands.install_locked_dependencies(pre, dev)` and `install_locked_dependencies` should be changed otherwise the locked dev dependencies will not be installed despite being locked. 

I saw #199 more about naming and conventions. Adding the ability to install `--dev` dependencies is separate I think... There is disambiguation to be done between "development (symlink)" installs and "dev (dev dependencies)" for sure.

PR #217, opened 2021-02-08
cli: separate vocabulary creation from demo record creation

max-moser:
closes inveniosoftware/invenio-rdm-records#347

* allows the cration of required vocabularies without creating demo records

this will likely require a dependency bump for `invenio-rdm-records` somewhere (`invenio-app-rdm`, and `cookiecutter-invenio-rdm`?)
otherwise, `invenio-cli setup` will fail due to the command `invenio rdm-records vocabularies` being unknown

### DIFF
--- a/invenio_cli/commands/containers.py
+++ b/invenio_cli/commands/containers.py
@@ -159,6 +159,21 @@ class ContainersCommands(ServicesCommands):
 
         return steps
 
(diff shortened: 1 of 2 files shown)


fenekku (approved):
We will bump versions later and the dependencies should be fine. :+1:

PR #218, opened 2021-02-19
commands: create admin user on setup

max-moser:
* the latest release of Invenio-RDM-Records requires the owners
  specified in each record's access.owned_by to exist
* additionally, at least one owner has to be specified for each record
* thus, the creation of demo records requires a user to exist
* to fix this, we create an admin user on setup
* the admin user's email address is based on the cookiecutter inputs


max-moser:
closed in favour of https://github.com/inveniosoftware/invenio-rdm-records/pull/395 and https://github.com/inveniosoftware/invenio-cli/pull/219, because it's planned that in the future, the fixtures from invenio-rdm-records take care of all the prerequisites

PR #225, opened 2021-03-30
Link inveniordm guide for setting up an environment to work on the module itself

avivace:
README only has usage instructions, but there are no mentions on how to correctly contribute on the tool itself. This PR should fix this by linking to the "Tools" paragraph of the invenio rdm maintainers guide that explains how to do this

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -2,7 +2,7 @@ Contributing
 ============
 
 Contributions are welcome, and they are greatly appreciated! Every
-little bit helps, and credit will always be given.
+little bit helps, and credit will always be given. Additional documentation can be found in the Invenio `maintainers guide <https://inveniordm.docs.cern.ch/maintenance/newcomers/#tools>`_.
 
 Types of Contributions
 ----------------------


README.rst
```diff
@@ -84,6 +84,12 @@ Containerized 'Production' environment
     $ invenio-cli containerize --no-install-js
 
 
+Development
+============
+
+To setup the environment for developing on `invenio-cli` itself, check out `this <https://inveniordm.docs.cern.ch/maintenance/newcomers/#tools>`_ paragraph on the InvenioRDM guide for maintainers.
+
```
fenekku, line 91: What about a link (in addition or instead) to the `CONTRIBUTING.md` file in this repo? It seems still valid... :thinking:

fenekku (approved):

README.rst
avivace: Makes sense, thanks @fenekku !
I'll move the link the `CONTRIBUTING.md` and mention the additional run-tests usage scenario

fenekku (approved):
Ah got it... What I had in mind was a link to the `CONTRIBUTING.rst` from the `README.md` (and then that added link). but that's a personal preference (and others might see it as obvious). This is fine.

Issue #196, opened 2020-11-30
FileNotFound exception when docker-compose is not found by check-requirements


avivace:
Package version (if known): 0.19.3

## Describe the bug

Hi! While following @ppanero intro to invenio-cli/invenio DRM, I noticed that when `docker-compose` is not found (but docker is there), a `FileNotFoundError` exception is thrown:

```
[...]
  File "/home/avivace/.local/share/virtualenvs/invenio_test-SmrCFEOj/lib/python3.9/site-packages/invenio_cli/helpers/process.py
", line 33, in run_cmd         
    p = popen(command, stdout=PIPE, stderr=PIPE)
  File "/usr/lib/python3.9/subprocess.py", line 947, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/usr/lib/python3.9/subprocess.py", line 1819, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'docker-compose'
```

## Steps to Reproduce

- Run `invenio-cli check-requirements` on a system without `docker-compose` available in the PATH.

## Expected behavior

An actual error message about docker-compose not being available.

ppanero:
Nice Thanks! We are checking versions... guess we assumed the installation, will fix that!

ppanero:
The existence of `node`, `docker` and `docker-compose` should be checked before the version check. For example, with `which`:

```
~ which docker
/usr/local/bin/docker
~ which not_exist
not_exist not found
```

Could be "greped" by `not found`.

PR #226, opened 2021-03-31
Show proper errors when docker and docker-compose executables are not found

avivace:
Fixes https://github.com/inveniosoftware/invenio-cli/issues/196. A couple of logging messages were also fixed to report the correct name of the executables being checked.

Example of the previous behaviour when the `docker-compose` executable was not found in the path: See https://github.com/inveniosoftware/invenio-cli/issues/196.

New behaviour:

image

This was applied to the `docker` and `docker-compose` version check steps but could be extended also to the other ones (e.g. node).


fenekku (approved):
Nice!

PR #242, opened 2021-05-10
invenio-cli: 

Pineirin:
implement cli commands for update and upgrade functionalities.
closes https://github.com/inveniosoftware/invenio-app-rdm/issues/657

### DIFF
--- a/invenio_cli/cli/packages.py
+++ b/invenio_cli/cli/packages.py
@@ -67,11 +67,26 @@ def install(cli_config, packages, skip_build):
 @pass_cli_config
 def outdated(cli_config):
     """Show outdated Python dependencies."""
-    raise click.UsageError("Not supported yet.")
+    steps = PackagesCommands.outdated_packages()
+
+    on_fail = "Some of the packages need to be updated."
+    on_success = "All packages are up to date."
+
+    run_steps(steps, on_fail, on_success)
 
 
 @packages.command()
+@click.argument("packages", nargs=-1, type=str)
 @pass_cli_config
-def update(cli_config):
-    """Update a single Python python package."""
-    raise click.UsageError("Not supported yet.")
+def update(cli_config, packages=None):
+    """Update all or some Python python packages."""
+    steps = PackagesCommands.update_packages(packages)
+
+    if packages:
+        on_fail = f"Failed to update packages {packages}."
+        on_success = f"Packages {packages} installed successfully."
+    else:
+        on_fail = "Failed to update packages."
+        on_success = "Packages installed successfully."
+
+    run_steps(steps, on_fail, on_success)
--- a/invenio_cli/commands/packages.py
+++ b/invenio_cli/commands/packages.py
@@ -38,6 +38,44 @@ class PackagesCommands(object):
         return steps
 
     @staticmethod
+    def outdated_packages():
+        """Steps to show outdated packages.
+
+        It is a class method since it does not require any configuration.
+        """
+        cmd = ['pipenv', 'update', '--outdated']
+
+        steps = [
+            CommandStep(
+                cmd=cmd,
+                env={'PIPENV_VERBOSITY': "-1"},
+                message="Checking outdated packages..."
+            )
+        ]
+
+        return steps
+
+    @staticmethod
+    def update_packages(packages):
+        """Steps to update all or some Python packages.
+
+        It is a class method since it does not require any configuration.
+        """
+        cmd = ['pipenv', 'update']
+        if packages:
+            cmd.extend(packages)
+
+        steps = [
+            CommandStep(
+                cmd=cmd,
+                env={'PIPENV_VERBOSITY': "-1"},
+                message="Updating package(s)..."
+            )
+        ]
+
+        return steps
+
+    @staticmethod
     def install_locked_dependencies(pre):
         """Installed dependencies from Pipfile.lock using sync."""
         # NOTE: sync has not interactive process feedback


invenio_cli/cli/cli.py
```diff
@@ -141,4 +141,15 @@ def destroy(cli_config):
 @invenio_cli.command()
 def upgrade():
     """Upgrades the current application to the specified newer version."""
-    raise click.UsageError("Not supported yet.")
+    os.popen('pipenv run invenio alembic upgrade')
```
ppanero, line 144: major: I am not sure I understand this...
- There are two functions with the same code inside, what's the reason behind it?
- The `upgrade` function is in the correct place... However, it should be done in the same fashion than the packages cli you implemented (e.g. `outdated`). This means, create a function at `command` level that return `CommandStep`s and son on... It should not invoke any `os` function at this level.

invenio_cli/cli/packages.py
```diff
@@ -9,6 +9,7 @@
 
 
 import click
+import os
```
ppanero, line 12: moderate: not needed?

invenio_cli/cli/packages.py
```diff
@@ -67,11 +68,28 @@ def install(cli_config, packages, skip_build):
...
     """Show outdated Python dependencies."""
-    raise click.UsageError("Not supported yet.")
+    steps = PackagesCommands.outdated_packages(packages)
+
+    on_fail = f"Failed to show outdated packages."
+    on_success = f"Outdated packages."
+
+    run_steps(steps, on_fail, on_success)
+
+    os.popen('pipenv update --outdated')
```
ppanero, line 78: major: This is something left out right? The command is being executre in the `run_steps`... if not, it should 😅

invenio_cli/cli/packages.py
```diff
@@ -67,11 +68,28 @@ def install(cli_config, packages, skip_build):
 @pass_cli_config
 def outdated(cli_config):
     """Show outdated Python dependencies."""
-    raise click.UsageError("Not supported yet.")
+    steps = PackagesCommands.outdated_packages(packages)
+
+    on_fail = f"Failed to show outdated packages."
+    on_success = f"Outdated packages."
```
ppanero, line 74: moderate: There is no need to use an `f`-string here. These are used for formating, i.e. when you need to add a variable value in the string e.g.

```python
example = "woow"
test = f"This is a {example} test"
```

Would result in `This is a woow test`. However, in this case, there is no variables.

```suggestion
    on_fail = "Failed to show outdated packages."
    on_success = "Outdated packages."
```

invenio_cli/cli/packages.py
```diff
@@ -67,11 +68,28 @@ def install(cli_config, packages, skip_build):
...
+
+    run_steps(steps, on_fail, on_success)
+
+    os.popen('pipenv update --outdated')
 
 
 @packages.command()
+@click.argument("package", nargs=1, type=str)
```
ppanero, line 82: doubt: have you tested this? I think it will result in a *mandatory* package argument, while we want it to be optional.

invenio_cli/commands/packages.py
```diff
@@ -37,6 +37,44 @@ def install_packages(packages):
...
+        It is a class method since it does not require any configuration.
+        """
+        prefix = ['pipenv', 'update']
+        cmd = prefix + ['--outdated']
+
+        steps = [
+            CommandStep(
+                cmd=cmd,
+                env={'PIPENV_VERBOSITY': "-1"},
+                message="Installing python dependencies..."
```
ppanero, line 53: moderate: Is this message correct? Is not installing anything, right? It's calculating outdated dependencies...

invenio_cli/cli/packages.py
```diff
@@ -67,11 +68,28 @@ def install(cli_config, packages, skip_build):
...
 @pass_cli_config
-def update(cli_config):
+def update(cli_config, package):
     """Update a single Python python package."""
```
ppanero, line 85: doubt: Can it update *all* outdated if `package` is not passed?

invenio_cli/cli/packages.py
```diff
@@ -67,11 +68,28 @@ def install(cli_config, packages, skip_build):
...
-    raise click.UsageError("Not supported yet.")
+    steps = PackagesCommands.update_packages(package)
+
+    if package not None:
+        on_fail = f"Failed to update package {package}."
+        on_success = f"Package {package} installed successfully."
+    else:
+        on_fail = f"Failed to update packages."
+        on_success = f"Packages installed successfully."
```
ppanero, line 93: minor: `f`-string not needed

invenio_cli/commands/packages.py
```diff
@@ -37,6 +37,44 @@ def install_packages(packages):
...
+        It is a class method since it does not require any configuration.
+        """
+        prefix = ['pipenv', 'update']
+        cmd = prefix + [package] if package not None else prefix
+
+        steps = [
+            CommandStep(
+                cmd=cmd,
+                env={'PIPENV_VERBOSITY': "-1"},
+                message="Installing python dependencies..."
```
ppanero, line 72: minor: In this case it is installing, but maybe specify that is "updating"?
```suggestion
                message="Updating python dependencies..."
```

invenio_cli/cli/cli.py
Pineirin: I forgot to rollback this, but the upgrade wasn't implemented in this draft. It wast just old code.

invenio_cli/cli/packages.py
Pineirin: I'm not sure I will have to check

invenio_cli/cli/packages.py
```diff
@@ -67,11 +67,26 @@ def install(cli_config, packages, skip_build):
 @pass_cli_config
 def outdated(cli_config):
     """Show outdated Python dependencies."""
-    raise click.UsageError("Not supported yet.")
+    steps = PackagesCommands.outdated_packages()
 
+    on_fail = ""
```
ppanero, line 72: major: This two string should contain a value... otherwise the user will not understand....

invenio_cli/cli/packages.py
```diff
@@ -67,11 +67,26 @@ def install(cli_config, packages, skip_build):
...
 
 @packages.command()
+@click.argument("package", nargs=-1, type=str)
 @pass_cli_config
-def update(cli_config):
+def update(cli_config, package):
     """Update a single Python python package."""
-    raise click.UsageError("Not supported yet.")
+
+    if len(package) == 1:
```
ppanero, line 83: moderate: i'm not sure the logic is correct... Firstly, does `pipenv` accept more than one package to update? i.e. `pipenv update package1 package2`?

If yes:
- It should do some processing to pass them all (e.g. `" ".join(packages)`)
If no:
- It should check that maximum of one package name is passed

Right now, what would happen is that if the user passes several packages, e.g.
`invenio-cli update invenio-app-rdm invenio-rdm-records`, it will update **all** and it might not be desirable.

invenio_cli/commands/packages.py
```diff
@@ -37,6 +37,44 @@ def install_packages(packages):
...
+                cmd=cmd,
+                env={'PIPENV_VERBOSITY': "-1"},
+                message="Checking outdated packages..."
+            )
+        ]
+
+        return steps
+
+    @staticmethod
+    def update_packages(package):
```
ppanero, line 60: naming: if the comment above results that more than one package can be updated at the same time... what about calling the parameter `packages`?

invenio_cli/commands/packages.py
```diff
@@ -37,6 +37,44 @@ def install_packages(packages):
...
+        """Steps to update all or one Python packages.
+
+        It is a class method since it does not require any configuration.
+        """
+        prefix = ['pipenv', 'update']
+        cmd = prefix + [package] if (package is not "") else prefix
```
ppanero, line 66: check it in the python console, but I'd swear that `""` string would evaluate to `False`. Meaning this could be changed to:
```suggestion
        cmd = prefix + [package] if not package else prefix
```

invenio_cli/commands/packages.py
ppanero, line 60: moderate: you can make the parameter optional:
```suggestion
    def update_packages(package=None):
```

That way above, when calling you do not need to pass `""`, in line *invenio_cli/cli/packages.py:88* would be:
```
steps = PackagesCommands.update_packages()
```

invenio_cli/cli/packages.py
Pineirin: The string would be redundant with the information already displayed by the command itself, it would be redundant.

invenio_cli/cli/packages.py
```diff
@@ -67,11 +67,26 @@ def install(cli_config, packages, skip_build):
...
-    """Update a single Python python package."""
-    raise click.UsageError("Not supported yet.")
+def update(cli_config, packages=[]):
+    """Update all or some Python python packages."""
+    if len(packages) >= 1:
+        steps = PackagesCommands.update_packages(packages)
+        on_fail = f"Failed to update packages {packages}."
+        on_success = f"Packages {packages} installed successfully."
+    else:
+        steps = PackagesCommands.update_packages("")
```
ppanero, line 88: major: this is not a good practice... either:
- The parameter is *mandatory* as it is now, then you **MUST** pass something meaningfull (or `None`)... as commented [here](https://github.com/inveniosoftware/invenio-cli/pull/242#discussion_r629384174). Passing `""` is too hackish.
- The parameter is *optional*, you deal with it in the function itself as mentioned [here](https://github.com/inveniosoftware/invenio-cli/pull/242#discussion_r629382704). Then you can simply call the function (`PackagesCommands.update_packages()`)

invenio_cli/cli/packages.py
```diff
@@ -67,11 +67,26 @@ def install(cli_config, packages, skip_build):
...
+    run_steps(steps, on_fail, on_success)
 
 
 @packages.command()
+@click.argument("packages", nargs=-1, type=str)
 @pass_cli_config
-def update(cli_config):
-    """Update a single Python python package."""
-    raise click.UsageError("Not supported yet.")
+def update(cli_config, packages=[]):
```
ppanero, line 81: major: never use `{}` or `[]` as default in parameters... They are *mutable* and can lead to unexpected errors...

See this https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil/

invenio_cli/commands/packages.py
```diff
@@ -37,6 +37,45 @@ def install_packages(packages):
...
+
+    @staticmethod
+    def update_packages(packages):
+        """Steps to update all or some Python packages.
+
+        It is a class method since it does not require any configuration.
+        """
+        cmd = ['pipenv', 'update']
+        for package in packages:
+            cmd.extend([package])
```
ppanero, line 67: hmm why do you need to loop and extend one by one? can't it be done directly as `cmd.extend(packages)`?

In any case, after the changes requested above you will need to check if for packages:

```python
if packages:
    cmd.extend(packages)
```

ppanero (commented):
outdated is good, update needs a bit of work still

invenio_cli/cli/packages.py
ppanero: I understand, but is more future proof (i.e. in case it becomes more than one step)...

invenio_cli/cli/packages.py
```diff
@@ -67,11 +67,26 @@ def install(cli_config, packages, skip_build):
...
+    steps = PackagesCommands.outdated_packages()
+
+    on_fail = "Some of the packages need to be updated."
+    on_success = "All packages are up to date."
+
+    run_steps(steps, on_fail, on_success)
 
 
 @packages.command()
+@click.argument("packages", nargs=-1, type=str)
```
ppanero, line 79: Is this optional? or is it forcing at least one? Does `argument` allow for defaults? (i.e. [])

invenio_cli/cli/packages.py
```diff
@@ -67,11 +67,26 @@ def install(cli_config, packages, skip_build):
...
 @pass_cli_config
-def update(cli_config):
-    """Update a single Python python package."""
-    raise click.UsageError("Not supported yet.")
+def update(cli_config, packages=None):
+    """Update all or some Python python packages."""
+    if len(packages) >= 1:
```
ppanero, line 83: if above does not allow defaults... you need to check for `if packages and len(pacakges) >= 1` otherwise, if packages is `None` it will throw an exception... `TypeError: object of type 'NoneType' has no len()`

invenio_cli/cli/packages.py
ppanero, line 79: Is this optional? or is it forcing at least one? Does argument allow for defaults? (i.e. [])

invenio_cli/commands/packages.py
```diff
@@ -17,7 +17,7 @@ class PackagesCommands(object):
     """Local installation commands."""
 
     @staticmethod
-    def install_packages(packages):
+    def install_packages(packages=None):
```
ppanero, line 20: major: why this change!!? install requires a minimum of 1 package, is different from update.

invenio_cli/commands/packages.py
```diff
@@ -37,6 +37,47 @@ def install_packages(packages):
...
+    def update_packages(packages=None):
+        """Steps to update all or some Python packages.
+
+        It is a class method since it does not require any configuration.
+        """
+        cmd = ['pipenv', 'update']
+        if packages:
+            cmd.extend(packages)
+
+        breakpoint()
```
ppanero, line 69: major: be careful ;)

invenio_cli/cli/packages.py
Pineirin: This can be controlled with the nargs parameter. If it is set to -1, then an unlimited number of arguments is accepted.

The value is then passed as a tuple. Note that only one argument can be set to nargs=-1, as it will eat up all arguments.

As far as I understand, it's optional, and testing it seems to work as intended.

ppanero: Sounds good :) I have made a tiny tiny change to avoid a few lines of code and argument passing :)

ppanero (approved):
LGTM! 🚀 nice job! Testing shows:

- `packages outdated command` ✅ 
- `packages update <list_of_pkgs>` ☑️... There is a limitation in terms of what can be updated *specifically*, and those are only packages present in the Pipfile. More details below...

```
$ invenio-cli packages outdated

Checking outdated packages...
Building requirements...
Resolving dependencies...
✔ Success!
Building requirements...
Resolving dependencies...
✔ Success!
Skipped Update of Package zipp: 3.4.1 installed,, 3.4.1 available.
Skipped Update of Package xmltodict: 0.12.0 installed,, 0.12.0 available.
[...]
Package 'testpath' out-of-date: '==0.4.4' installed, '==0.5.0' available.
Package 'marshmallow' out-of-date: '==3.12.0' installed, '==3.12.1' available.
Package 'MarkupSafe' out-of-date: '==1.1.1' installed, '==2.0.0' available.
Package 'invenio' out-of-date: '==3.4.0' installed, '==3.4.1' available.
Package 'invenio-userprofiles' out-of-date: '==1.2.1' installed, '==1.2.2' available.
Package 'invenio-records' out-of-date: '==1.5.0a5' installed, '==1.5.0a7' available.
Package 'invenio-base' out-of-date: '==1.2.3' installed, '==1.2.4' available.
Package 'invenio-assets' out-of-date: '==1.2.5' installed, '==1.2.6' available.
Package 'invenio-accounts' out-of-date: '==1.4.3' installed, '==1.4.4' available.
Package 'Flask' out-of-date: '==1.1.2' installed, '==1.1.4' available.
Package 'Flask-RESTful' out-of-date: '==0.3.8' installed, '==0.3.9' available.
Package 'Faker' out-of-date: '==8.1.2' installed, '==8.1.4' available.
Package 'decorator' out-of-date: '==5.0.7' installed, '==5.0.9' available.
Some of the packages need to be updated.
```

Then if you try to update some not present in the Pipfile will fail, which means we cannot update a specific invenio package without upgrading the full `invenio`. This could eventually be useful to make users go for bundled versions and not start deviating from the norm and having to calculate compatibilities.

```
$ invenio-cli packages update invenio-base invenio-records
Updating package(s)...
Warning: invenio-base was not found in your Pipfile! Aborting.
Aborted!
Failed to update packages ('invenio-base', 'invenio-records').
```

Full invenio update, works. Note **it does not update the Pipfile, only the Pipfile.lock** (outputs shortened)

```
$ invenio-cli packages update invenio
Updating package(s)...
Locking [dev-packages] dependencies...
Building requirements...
Resolving dependencies...
✔ Success!
Locking [packages] dependencies...
Building requirements...
Resolving dependencies...
✔ Success!
Updated Pipfile.lock (09210d)!
Installing dependencies from Pipfile.lock (09210d)...
  🐍   ▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉ 13/13 — 00:00:05
To activate this project's virtualenv, run pipenv shell.
Alternatively, run a command inside the virtualenv with pipenv run.
All dependencies are now up-to-date!
Packages ('invenio',) installed successfully.

$ cat Pipfile

[packages]
invenio = "~=3.4.0"
[...]

$ pip freeze | grep invenio
invenio==3.4.1
[...]

$ In the Pipfile.lock

        "invenio": {
            "extras": [
                "auth",
                "base",
                "files",
                "metadata"
            ],
            "hashes": [
                "sha256:d480e1847c681220624563c28f44a0bcc24eb7457e80b9594ff1c285c86bd69b",
                "sha256:eddc4d9e8f0f8a1e2efada5a582c9b45d9fe3e49740051e2fe0fe8ec999f94f0"
            ],
            "index": "pypi",
            "version": "==3.4.1"
        },
```

After the update, it seems all packages are up to date... so a simple `invenio-cli packages update invenio` did an update of everything (due to chaining deps)

```
$ invenio-cli packages outdated

Checking outdated packages...
Building requirements...
Resolving dependencies...
✔ Success!
Building requirements...
Resolving dependencies...
✔ Success!
[...]
All packages are up to date!
All packages are up to date.
```

PR #246, opened 2021-05-19
global: add command to update invenio-app-rdm package

Pineirin:
Include a command to update invenio-app-rdm.
closes https://github.com/inveniosoftware/invenio-app-rdm/issues/849

### DIFF
--- a/invenio_cli/cli/packages.py
+++ b/invenio_cli/cli/packages.py
@@ -76,16 +76,19 @@ def outdated(cli_config):
 
 
 @packages.command()
-@click.argument("packages", nargs=-1, type=str)
+@click.argument("version", required=False, type=str)
 @pass_cli_config
-def update(cli_config, packages=None):
+def update(cli_config, version=None):
     """Update all or some Python python packages."""
-    steps = PackagesCommands.update_packages(packages)
-
-    if packages:
-        on_fail = f"Failed to update packages {packages}."
-        on_success = f"Packages {packages} installed successfully."
+    if version:
+        db = cli_config.get_db_type()
+        es = f"elasticsearch{cli_config.get_es_version()}"
+        package = f"invenio-app-rdm[{db},{es}]~="
+        steps = PackagesCommands.update_package_new_version(package, version)
+        on_fail = f"Failed to update version {version}"
+        on_success = f"Version {version} installed successfully."
     else:
+        steps = PackagesCommands.update_packages()
         on_fail = "Failed to update packages."
         on_success = "Packages installed successfully."
 
--- a/invenio_cli/commands/packages.py
+++ b/invenio_cli/commands/packages.py
@@ -56,20 +56,37 @@ class PackagesCommands(object):
         return steps
 
     @staticmethod
-    def update_packages(packages):
-        """Steps to update all or some Python packages.
+    def update_packages():
+        """Steps to update all Python packages.
 
         It is a class method since it does not require any configuration.
         """
         cmd = ['pipenv', 'update']
-        if packages:
-            cmd.extend(packages)
 
         steps = [
             CommandStep(
                 cmd=cmd,
                 env={'PIPENV_VERBOSITY': "-1"},
                 message="Updating package(s)..."
+            )
+        ]
+
+        return steps
+
+    @staticmethod
+    def update_package_new_version(package, version):
+        """Update invenio-app-rdm version.
+
+        It is a class method since it does not require any configuration.
+        """
+        prefix = ['pipenv']
+        app = prefix + ['install', package + version]
+
+        steps = [
+            CommandStep(
+                cmd=app,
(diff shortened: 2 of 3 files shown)


invenio_cli/cli/cli.py
```diff
@@ -146,3 +146,13 @@ def upgrade():
     on_success = "Upgrade sucessfull."
 
     run_steps(steps, on_fail, on_success)
+
+
+@invenio_cli.command()
+def updateapp():
+    """Upgrades the version of invenio-app-rdm 3.0.0."""
```
fenekku, line 153:
```suggestion
    """Updates the dependencies of your instance without upgrading to 4.0.0."""
```

invenio_cli/commands/upgrade.py
```diff
@@ -52,3 +52,32 @@ def upgrade():
         ]
 
         return steps
+
+    @staticmethod
+    def update_app():
+        """Upgrades the version of invenio-app-rdm 3.0.0.
```
fenekku, line 58:
```suggestion
        """Update dependencies..
```

invenio_cli/commands/upgrade.py
```diff
@@ -52,3 +52,32 @@ def upgrade():
...
+
+        It is a class method since it does not require any configuration.
+        """
+        prefix = ['pipenv']
+        app = prefix + [
+            'install', 'invenio-app-rdm[postgresql, elasticsearch7]~=3.0.0'
```
fenekku, line 64: Just to account for possible minor updates:

```suggestion
            'install', 'invenio-app-rdm[postgresql, elasticsearch7]~=3.0'
```
(i.e. this allows update to 3.1.0 if ever we do release that, while not upgrading to 4.0 if it exists)

Overall it does tie the invenio-cli code with the instance version i.e. 
- for invenio-cli developers, we need to update this piece of code for each release
- for administrators, upgrading invenio-cli would cause this command to upgrade their instance perhaps unwillingly

In the future we could parse the Pipfile and use the version constraints for invenio-app-rdm found there to circumvent that problem. Make an issue for this if not tackled now.

fenekku (commented):
Some minor things + discussion about using Pipfile's requirement instead.

invenio_cli/commands/upgrade.py
```diff
@@ -52,3 +52,33 @@ def upgrade():
...
+        It is a class method since it does not require any configuration.
+        """
+        prefix = ['pipenv']
+        app = prefix + [
+            'install',
+            'invenio-app-rdm[postgresql, elasticsearch7]~=' + version
+        ]
+        dependencies = prefix + [
+            'update'
+        ]
```
fenekku, line 69: Running `pipenv install invenio-app-rdm[postgresql, elasticsearch7]~=<nvew version>` and then `pipenv update` gave the same result as just running `pipenv install invenio-app-rdm[postgresql, elasticsearch7]~=<nvew version>` only. Am I missing something or is `pipenv update` redundant here?

invenio_cli/cli/cli.py
```diff
@@ -146,3 +146,14 @@ def upgrade():
     on_success = "Upgrade sucessfull."
 
     run_steps(steps, on_fail, on_success)
+
+
+@click.argument("version", nargs=1, type=str)
+@invenio_cli.command('update-app')
+def update_app(version):
+    """Updates the dependencies of your instance without upgrading to 4.0.0."""
```
fenekku, line 154: Now that we pass the version, we can remove reference to an explicit version in the helper text.

```suggestion
    """Updates the dependencies of your instance without upgrading to the next major version."""
```

fenekku (commented):
When @lnielsen mentioned passing the version, I understood it as passing it for the `invenio-cli upgrade <version>` command. For the `invenio-cli update-app` (why didn't we use the name `update`in the end?) command, the version we need to not go over (e.g. `~=3.0`) is already in the Pipfile and pipenv will respect it when doing `pipenv update`. So we just need to run `pipenv update` under the hood for `invenio-cli update-app`, but we want to change `invenio-cli upgrade` to take a version. @lnielsen can clarify if I am off track here.

invenio_cli/commands/upgrade.py
Pineirin: I thought that pipenv update was needed to keep all the dependencies updated, I will remove that command.

lnielsen:
So my take woudl be:

```
# execute locally
invenio-cli packages update # get latest patch-level releases
invenio-cli packages update v4.0 # upgrade invenio-app-rdm to from vX to v4.0 and get all latest patch-level releases

# execute on each instances (local, sandbox, prod)
invenio-cli upgrade   # alembic upgrade, data migration, rebuild all indexes
```

Pineirin (commented):
Changes seems alright to me, even do I have to test the behavior of including the package when updating a specific version.

PR #248, opened 2021-05-25
Requirements

Pineirin:
Add check requirement step for ImageMagick.

### DIFF
--- a/invenio_cli/commands/requirements.py
+++ b/invenio_cli/commands/requirements.py
@@ -120,6 +120,24 @@ class RequirementsCommands(object):
             )
 
     @classmethod
+    def check_imagemagick_version(cls, major, minor=-1, patch=-1,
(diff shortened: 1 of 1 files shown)


Pineirin:
I don't know how do display it as warning, there is a warning parameter, which does nothing. I would fix that part of the code, but I would like to consult that decision.
Sorry for the late commit, I had to stop early today.

ppanero (approved):
LGTM, I think for now is not worth putting effort into making "warnings" we can simply hard fail here (what is actually happening, right?)

Running check-requirements does not avoid the user to not run the instance, and its the last check so everything else (docker, python, etc.) is checked before... I would say we keep it :)

Nice job!

ppanero:
warning issue will be tackled in another task: https://github.com/inveniosoftware/invenio-cli/issues/249

PR #251, opened 2021-05-27
requirements

Pineirin:
Add check requirement step for ImageMagick.

### DIFF
--- a/invenio_cli/commands/requirements.py
+++ b/invenio_cli/commands/requirements.py
@@ -125,7 +125,7 @@ class RequirementsCommands(object):
         """Check the ImageMagick version."""
         # Output comes in the form of 'ImageMagick, version 7.0.11-13\n'
         try:
-            result = run_cmd(["magick", "-version"])
+            result = run_cmd(["convert", "--version"])
             version = cls._version_from_string(result.output.strip())
             return cls._check_version(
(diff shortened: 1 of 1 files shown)


invenio_cli/commands/requirements.py
```diff
@@ -125,12 +127,13 @@ def check_imagemagick_version(cls, major, minor=-1, patch=-1,
         """Check the ImageMagick version."""
         # Output comes in the form of 'ImageMagick, version 7.0.11-13\n'
         try:
-            result = run_cmd(["magick", "-version"])
-            version = cls._version_from_string(result.output.strip())
-            return cls._check_version(
-                "ImageMagick", version, major, minor, patch, exact
+            get_distribution('wand')
```
lnielsen, line 130: Hmm.....there's a chicken-egg problem here. This will run in the Invenio-CLI python process where we normally don't install Wand, so it will fail. Now, we run ``check-requirements`` before doing an install, which means that also in the pipenv you wont' have Wand installed. So seems like we need a better solution for detecting if imagemagick is installed.

lnielsen: I would give it a try with the ``convert`` command instead

lnielsen: So like before just change ``magick`` to ``convert``

PR #256, opened 2021-07-15
install: fix result error check on update instance path

zzacharo:


invenio_cli/commands/install.py
```diff
@@ -42,9 +42,9 @@ def update_instance_path(self):
             ['pipenv', 'run', 'invenio', 'shell', '--no-term-title',
                 '-c', '"print(app.instance_path, end=\'\')"']
         )
-        self.cli_config.update_instance_path(result.output.strip())
-
-        result.output = "Instance path updated successfully."
+        if not result.error:
```
ppanero, line 45: Should we rely on the error message or on the `status_code`... if it is not changed to > 0 maybe we should change it?

zzacharo: it is changed to `1`...isn't it equivalent? maybe I can rely indeed on the status code only and let the error message just pop up in the console...

ppanero: If it is changed to `1` is fine, because the later code with show a warning/error... But I would rely in the `status_code` as a practice... It might in general a command might not have an error message but still fail.

```python
 if results.status_code == 0:
```

PR #275, opened 2021-11-01
fix check requirements

ppanero:
Not sure what's the current board so I asked for review :)


fenekku (approved):
That does it :slightly_smiling_face:

PR #276, opened 2021-12-07
config: remove need for pre-existing .invenio.private per team member

### LINKED ISSUE #130: Avoid usage of ".invenio.private" when possible
Storing values in `.invenio.private` should be avoided whenever possible. The location of the `.invenio` file is usually enough to be able to compute all the necessary values at the beginning of running the `invenio-cli` command (similar to how `pipenv` treats the `Pipfile` as a point of reference):

- `project_dir`: this is the absolute path of the `.invenio` file
- `instance_path`: Since `.invenio` is in the same directory as the `Pipfile`, using `pipenv --venv` would return the absolute virtualenv path (even in the case that one was already created and activated from before). We can then append `var/instance` to this value

Note: this is not meant to be a full removal of `.invenio.private`, which should be still be used for hardcoding values. The incentive is to minimize local state/files that need to be gitignore'd and to provide sane defaults for users of the CLI.

fenekku:
closes #130

To summarize: as part of a team I want to clone the instance repository and run `invenio-cli install` ... without having to artificially create a `.invenio.private` with a path to a virtualenv that I also need to create. This had to be done before because these steps were done by `invenio-cli init` which isn't run because my teammate already ran it (I am cloning the result of that command afterall). This is also handy even without teammates, if one is cloning the repo on a different computer.  

This PR addresses that by relying on the in-memory project directory directly (no need to look-up the `.invenio.private` file) and the fact that `invenio-cli install` sets the instance location. Even though we technically still rely on .invenio.private for it, it should really be thought as a cache that is primed before use by `invenio-cli install`.

### DIFF
--- a/invenio_cli/cli/cli.py
+++ b/invenio_cli/cli/cli.py
@@ -23,7 +23,7 @@ from .containers import containers
 from .install import install
 from .packages import packages
 from .services import services
-from .utils import calculate_instance_path, pass_cli_config, run_steps
+from .utils import pass_cli_config, run_steps
 
 
 @click.group()
@@ -88,8 +88,7 @@ def init(flavour, template, checkout):
 
         click.secho("Writing invenio-invenio_cli config file...", fg='green')
         saved_replay = cookiecutter_wrapper.get_replay()
-        instance_path = calculate_instance_path(project_dir)
-        CLIConfig.write(project_dir, flavour, saved_replay, instance_path)
+        CLIConfig.write(project_dir, flavour, saved_replay)
 
         click.secho("Creating logs directory...", fg='green')
         os.mkdir(Path(project_dir) / "logs")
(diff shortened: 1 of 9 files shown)


invenio_cli/cli/cli.py
```diff
@@ -88,8 +88,7 @@ def init(flavour, template, checkout):
 
         click.secho("Writing invenio-invenio_cli config file...", fg='green')
         saved_replay = cookiecutter_wrapper.get_replay()
-        instance_path = calculate_instance_path(project_dir)
```
fenekku, line 91: This was actually always returning an error that was hidden away by our `ProcessResponse`, so it doesn't even make sense to run it.

fenekku:
not ready yet

fenekku:
ready for review

PR #281, opened 2022-02-03
cli: add --no-input and --config options for unsupervised init (closes #280)

### LINKED ISSUE #280: Option to create an Invenio instance without user input (e.g. for use within CI)
I want to run blackbox tests against the backend API of Invenio RDM and would therefore like to have a possibility to create in instance within the CI pipeline to test against.

Currently, I am using an ugly workaround of creating instance using cookiecutter directly with `--no-input` and copying the `.invenio` file from a "validly" generated instance. This, obviously, is not a proper solution (and only works easily for the default values).

## Describe the solution you'd like
`invenio-cli init` should have a flag `--cookiecutter-config` that takes a JSON file with the values. It could be loaded and passed as a dict through the cookiecutter API as described [here](https://cookiecutter.readthedocs.io/en/1.7.0/advanced/suppressing_prompts.html).

## Additional context
If there is no opposition against this idea, I could also take care of implementing this little addition in the way outlined above. But first, I wanted to get other opinions.

apirogov:
closes https://github.com/inveniosoftware/invenio-cli/issues/280

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -3,6 +3,7 @@ MIT License
 Copyright (C) 2019 CERN.
 Copyright (C) 2019 Northwestern University.
 Copyright (C) 2021 TU Wien.
+Copyright (C) 2022 Forschungszentrum Jülich GmbH.
 
 Permission is hereby granted, free of charge, to any person obtaining a copy of
 this software and associated documentation files (the "Software"), to deal in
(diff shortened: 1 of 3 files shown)


lnielsen:
Looks good to me. Just want to take to it for a quick spin, so I'll try to merge tomorrow when I have time.

lnielsen (approved):

fenekku (approved):

fenekku:
Took it for a spin and it worked well. Thanks!

PR #287, opened 2022-04-04
Add support for `poetry` as an optional, alternative packaging backend

max-moser:
`pipenv` has a history of... being difficult at times.`poetry` is the newish kid on the block, and looks quite promising.
I've had scenarios where `poetry` was able to lock the dependencies within seconds or minutes, while `pipenv` couldn't come up with any results in an hour.
As such, it would be nice to have the option to work with `poetry` rather than `pipenv` (if you know what you're doing).

This PR adds support for `poetry` for most of the operations as an alternative packaging tool.
One known issue in `invenio-cli` is that `poetry show --outdated` will always exit with a status code of `0`, and thus `invenio-cli` will always report that all packages are up to date (even though the previous output clearly suggested the opposite).
Further, the handling of pre-release package versions is different between `pipenv` and `poetry`: It seems that `poetry` only accepts an `--allow-prereleases` flag when installing additional packages, but neither on locking nor installing dependencies from the lock file (which I feel makes sense).
Another notable difference is that `poetry` does not, by itself, read the environment variables from a `.env` file. This would have to be done in code, likely via [`python-dotenv`](https://pypi.org/project/python-dotenv/) before executing `poetry` commands.

**Note:** By default, `invenio-cli` will still use `pipenv`.
The configuration (`[cookiecutter]`) has to request `packaging_backend = poetry` specifically to switch to `poetry`.

### DIFF
--- a/invenio_cli/cli/assets.py
+++ b/invenio_cli/cli/assets.py
@@ -8,7 +8,6 @@
 """Invenio module to ease the creation and management of applications."""
 
 import click
-from click_default_group import DefaultGroup
 
 from ..commands import AssetsCommands
 from .utils import pass_cli_config, run_steps
--- a/invenio_cli/cli/cli.py
+++ b/invenio_cli/cli/cli.py
@@ -45,8 +45,15 @@ invenio_cli.add_command(services)
               help='Check development requirements.')
 def check_requirements(development):
     """Checks the system fulfills the pre-requirements."""
+    try:
+        # the 'check-requirements' command should probably be available
+        # from outside the project directory as well
+        cli_config = CLIConfig()
+    except FileNotFoundError:
+        cli_config = None
+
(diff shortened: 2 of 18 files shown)


max-moser:
funny, i was thinking about closing this PR just a few days back :D
but if there's still interest in this, i can have a further look

apirogov:
We'd be quite happy to beta-test the "experimental" poetry-based instance setup once this is merged :)

At least I have some more experience dealing with the quirks of poetry, while all of us don't have experience with pipenv, adding up to the layers of abstraction Invenio introduces and we are just learning to navigate... :sweat_smile:

max-moser:
@avivace @apirogov maybe it's not relevant for you anymore, but a few hours ago #384 got merged and released as v1.6.0, which enables `uv` as an optional python package manager!
some details are given in that PR.

PR #289, opened 2022-05-25
migrate setup and black

utnapischtim:
- migrate setup.py to setup.cfg
- migrate to use black as opinionated auto formater
- add .git-blame-ignore-revs
- global: clean test infrastructure


.github/workflows/tests.yml
```diff
@@ -20,25 +29,11 @@ jobs:
...
-          include:
-          - services: release
-            REQUIREMENTS: "release"
-            REQUIREMENTS_LEVEL: "pypi"
-          - services: lowest
-            REQUIREMENTS: "lowest"
-            REQUIREMENTS_LEVEL: "min"
-          - services: devel
-            REQUIREMENTS: "devel"
-            REQUIREMENTS_LEVEL: "dev"
```
kpsherva, line 36: I am not sure if you were planning to remove the devel and lowest, it could make sense to leave them, @lnielsen what do you think?

lnielsen: I would keep ``min`` (changing name from lowest), but remove ``devel``.

Issue #291, opened 2022-05-31
If `ImageMagick` isn't installed, `check-requirements` will report both failure and success


max-moser:
Package version (if known): 1.0.4

## Describe the bug
The output of `invenio-cli check-requirements` is confusing, because it mentions both success as well as failure.

## Steps to Reproduce
1. Install Ubuntu (e.g. on a VM)
2. Install most basic dependencies (but not `ImageMagick`... which is listed as an "extended" dependency in the docs: https://inveniordm.docs.cern.ch/install/requirements/)
3. Execute `invenio-cli check-requirements`

## Expected behavior

When one of the dependencies isn't found, no success message should be displayed.

Also, should `ImageMagick` even be checked without `--development` flag?

## Screenshots (if applicable)

![image](https://user-images.githubusercontent.com/6437519/171190053-2fcad11f-5da6-49e8-bdf7-01e026ba3ec0.png)

PR #292, opened 2022-05-31
Update `check-requirements` command

max-moser:
Closes https://github.com/inveniosoftware/invenio-cli/issues/291

This PR aims to make the `invenio-cli check-requirements [--development]` command a bit more consistent.

### DIFF
--- a/invenio_cli/commands/requirements.py
+++ b/invenio_cli/commands/requirements.py
@@ -205,7 +205,12 @@ class RequirementsCommands(object):
                 func=cls.check_npm_version,
                 args={"major": 6, "exact": True},
                 message="Checking NPM version..."
-            )
+            ),
+            FunctionStep(
+                func=cls.check_imagemagick_version,
+                args={"major": 0, "minor": 0},
+                message="Checking ImageMagick version...",
+            ),
         ]
 
         return steps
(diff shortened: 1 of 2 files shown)


invenio_cli/commands/requirements.py
```diff
@@ -148,22 +170,26 @@ def check_imagemagick_version(cls, major, minor=-1, patch=-1,
...
         """Check the pipenv version."""
         # Output comes in the form of 'pipenv, version 2020.11.15\n'
-        result = run_cmd(["pipenv", "--version"])
+        try:
+            result = run_cmd(["pipenv", "--version"])
+
+            parts = result.output.strip().split(',')
+            if parts[0] != 'pipenv':
+                # note that result.error is stderr (type str) of the process
+                raise Exception(result.error)
```
fenekku, line 179: It can be the case that pipenv --version returns a string that doesn't start with `pipenv`?

Why not the ProcessResponse here? (if the above is valid)
```
return ProcessResponse(
                error=f"Pipenv not ok. Got {parts[0]}.", status_code=1
            )
```

fenekku (approved):
Question around the pipenv checks but rest works (approving so not blocking)

invenio_cli/commands/requirements.py
max-moser: i think that could be the case if somehow `pipenv` points to a different binary, e.g. if it's aliased or symlinked or something... i just kept the original logic here.
regarding returning the `ProcessResponse`, that's probably a better idea than raising an `Exception` b/c it's more legible. will update.

fenekku: > points to a different binary
Sure maybe the original logic was a little too weary :shrug: ...

PR #298, opened 2022-06-24
Check for git as requirement

avivace:
Follows up https://github.com/inveniosoftware/docs-invenio-rdm/pull/379.

`invenio-cli check-requirements` now checks if git is installed, without paying attention to the version (similar to what we do with imagemagick).

I added the step in the `development` requirements, but I'm not super sure about it, since `git` is required even for the `invenio-cli init rdm -c <version>` command, which is suggested in both [installation methods](https://inveniordm.docs.cern.ch/install/), local and containers.

### DIFF
--- a/invenio_cli/commands/requirements.py
+++ b/invenio_cli/commands/requirements.py
@@ -151,6 +151,20 @@ class RequirementsCommands(object):
             )
 
     @classmethod
+    def check_git_version(cls, major, minor=-1, patch=-1, exact=False):
+        """Check the git version."""
+        # Output comes in the form of 'git version 2.36.1\n'
+        try:
+            result = run_cmd(["git", "--version"])
+            version = cls._version_from_string(result.output.strip())
+            return cls._check_version("git", version, major, minor, patch, exact)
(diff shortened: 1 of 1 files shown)


invenio_cli/commands/requirements.py
```diff
@@ -150,6 +150,20 @@ def check_imagemagick_version(cls, major, minor=-1, patch=-1, exact=False):
                 status_code=1,
             )
 
+    @classmethod
+    def check_git_version(cls, major, minor=-1, patch=-1, exact=False):
+        """Check the ImageMagick version."""
```
kpsherva, line 155: docstring does not match the method purpose

avivace: Ups, fixed it

ppanero (approved):
@avivace thanks for the contribution. I think is more than okay in development. while we do support the container setup I guess it does not make much sense to pull from master or other branch since that might involve installing local modules and so on which is not really supported/possible in containers. I see `-c master` as a development feature mostly.

PR #310, opened 2022-09-29
[DO NOT MERGE] services: make commands aware of rdm version

ppanero:
I do not like this way because it requires installing `invenio-app-rdm` which technically cli should not be aware of (i.e. in containerized version).

Shall we parse the `Pipfile`?


invenio_cli/commands/services.py
```diff
@@ -164,36 +165,43 @@ def _setup(self):
...
-                    "communities",
-                    "custom-fields",
-                    "init",
-                ],
-                env={"PIPENV_VERBOSITY": "-1"},
-                message="Creating custom fields for communities...",
-            ),
+        ]
+
+        if rdm_version[0] < 10:
```
ntarocco, line 170:
```suggestion
        if rdm_version[0] >= 10:
```
?

What if we do something more structured, e.g. create a enum/map of features with version, for example (allow me pseudo-code):
```
versions = { 
  10: [custom-fields, administration],
   ...
}
```
then an helper:
```
get_enabled_features():
  return versions[current-app-rdm version]
```

and then in the code we get that map and use it?
```
if custom-fields-in-version:
  run init command
```

ppanero:
@ntarocco I've for now just parsed the app-rdm version from the pipfile (I tried with the existing configparser but it doesnt work, my guess is the double [[ ]] at the beginning)

Regarding the more complex logic of commands/features per version I do agree but I think we can leave it for a different/PR issue and tackle it along with the distinction of RDM/ILS of the cli? so we unblock this on the release

invenio_cli/helpers/rdm.py
```diff
@@ -0,0 +1,24 @@
...
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""Invenio CLI dependencies helper."""
+
+import re
+
+from pipfile import Pipfile
+
+
+def rdm_version():
```
ntarocco, line 16: 👍🏼

ntarocco (approved):

ppanero:
Opened relevant issues:
- Make commands flavour agnostic https://github.com/inveniosoftware/invenio-cli/issues/311
- https://github.com/inveniosoftware/invenio-cli/issues/312

PR #319, opened 2022-10-12
services: fix wait for services to start

ntarocco:
The `click` context is injected in many places (e.g. `click.echo`) but it should instead be only used in the `cli` module, not in `commands` module. With current code, is quite easy to end up with circular imports and it is not easy to make the code testable.
Tip: avoid global context (`import click`) as much as possible where it is not needed.

Some funcs are aware of how they are called: they should not be. A func should have a clear API and the caller needs to take care of the response and take actions depending on the response.

### DIFF
--- a/invenio_cli/cli/cli.py
+++ b/invenio_cli/cli/cli.py
@@ -19,6 +19,7 @@ from ..commands import (
     ContainersCommands,
     LocalCommands,
     RequirementsCommands,
+    ServicesCommands,
     UpgradeCommands,
 )
 from ..helpers.cli_config import CLIConfig
@@ -28,7 +29,7 @@ from .containers import containers
 from .install import install
 from .packages import packages
 from .services import services
-from .utils import pass_cli_config, run_steps
+from .utils import handle_process_response, pass_cli_config, run_steps
 
 
 @click.group()
@@ -167,6 +168,12 @@ def run(cli_config, host, port, debug, services, celery_log_file):
 
     NOTE: this only makes sense locally so no --local option
     """
+    if services:
+        cmds = ServicesCommands(cli_config)
+        response = cmds.ensure_containers_running()
+        # fail and exit if containers are not running
+        handle_process_response(response)
+
     commands = LocalCommands(cli_config)
     commands.run(
         host=host,
--- a/invenio_cli/cli/services.py
+++ b/invenio_cli/cli/services.py
@@ -10,7 +10,7 @@
 
 import click
 
-from ..commands import Commands, ServicesCommands
+from ..commands import ServicesCommands
 from .utils import pass_cli_config, run_steps
 
 
@@ -70,7 +70,6 @@ def setup(cli_config, force, no_demo_data, stop_services, services):
     run_steps(steps, on_fail, on_success)
 
 
-# FIXME: Duplicated code from containers.py
 @services.command()
 @click.option(
     "-v",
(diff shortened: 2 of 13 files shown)


invenio_cli/cli/cli.py
```diff
@@ -167,6 +168,12 @@ def run(cli_config, host, port, debug, services, celery_log_file):
 
     NOTE: this only makes sense locally so no --local option
     """
+    if services:
```
ntarocco, line 171: Moved outside to avoid circual imports

invenio_cli/cli/utils.py
```diff
@@ -18,23 +18,36 @@ def run_steps(steps, fail_message, success_message):
...
-            click.secho(fail_message, fg="yellow")
-        elif result.output:
-            click.secho(message=result.output, fg="green")
+        response = step.execute()
+        handle_process_response(response, fail_message=fail_message)
     else:
         click.secho(message=success_message, fg="green")
+
+
+def handle_process_response(response, fail_message=None):
```
ntarocco, line 27: Extracted func to be reusable, independently of steps execution.

invenio_cli/commands/services_health.py
```diff
@@ -100,62 +98,71 @@ def redis_healthcheck(cls, *args, **kwargs):
...
-        of all the services.
-        """
-        if len(services) == 0:
-            services = HEALTHCHECKS.keys()
-
-        for service in services:
-            exp_backoff_time = 2
-            try_ = 1
-            check = HEALTHCHECKS[service]
-            ready = check(
```
ntarocco, line 125: The bug was here: `ready` is a `ProcessResponse` obj, and not a `boolean`.

invenio_cli/commands/services_health.py
```diff
@@ -100,62 +98,71 @@ def redis_healthcheck(cls, *args, **kwargs):
...
 
 
 HEALTHCHECKS = {
-    "search": ServicesHealthCommands.search_healthcheck,
-    "postgresql": ServicesHealthCommands.postgresql_healthcheck,
-    "mysql": ServicesHealthCommands.mysql_healthcheck,
-    "redis": ServicesHealthCommands.redis_healthcheck,
+    "search": {
+        "func": ServicesHealthCommands.search_healthcheck,
+        "initial_delay": 15,  # search cluster can be particularly slow to start
```
ntarocco, line 153: `OpenSearch` is particularly slow to start up: introduced an initial delay because even the exponential backoff was failing.

invenio_cli/cli/services.py
```diff
@@ -90,16 +90,19 @@ def status(cli_config, verbose):
...
 
-    messages = [
-        {"message": "{}: up and running.", "fg": "green"},
-        {"message": "{}: unable to connect or bad response.", "fg": "red"},
-        {"message": "{}: no healthcheck function defined.", "fg": "yellow"},
-    ]
-
-    for idx, status in enumerate(statuses):
-        message = messages[status]
+    messages = {
```
ppanero, line 93: minor/thinking out loud: these few lines of code are repeated in `containers.py`, and they are similar in objective to the `handle_process_response` function.

Can they be moved to a function? Maybe they are small enough to not need to, also _utils_ is not a good place to put things on (personal biased). So I'm not sure where to move it...

ppanero (approved):

invenio_cli/cli/services.py
ntarocco: Done, `containers` now invokes `setup` func `status`

PR #324, opened 2022-10-21
docker: generalize docker container name checking

zzacharo:


ntarocco (approved):

invenio_cli/helpers/docker_helper.py
```diff
@@ -43,6 +43,20 @@ def _normalize_name(self, project_shortname):
         else:
             return project_shortname
 
+    def _get_container_from_service(self, service_name):
+        """Retrieve the docker container for the given service_name."""
+        container = None
```
ppanero, line 48: minor: There is an outer scope `container` variable, which is only used for a return at the end. However, there is anotther `container` variable inside the for loop... Could we avoid the outer `container-=None` and just have at the end:

```python
if container_name:
    return self.docker_client...
    
return None
```

Or a one liner `return None if not container_name else self.docker_client...`

invenio_cli/helpers/docker_helper.py
```diff
@@ -100,15 +114,16 @@ def destroy_containers(self):
...
-            tty=True,
-            stdout=True,
-            stderr=True,
-        )
-        # FIXME: What happens when exec_run fails? handle exception.
-        return ProcessResponse(
-            output=status.output.decode("utf-8").strip(), status_code=status.exit_code
-        )
+        container = self._get_container_from_service("web-ui")
+        if container:
```
ppanero, line 118: moderate: should we add an `else` clause and "hack" a custom `ProcessResponse` with e.g. `status_code=1` and `output="Web UI container not found. Is it up and running?`. Otherwise it would silently fail if no container (Is it possible to reach this part of the code with the containers down?)

zzacharo: if you are following the high level api then containers should always be up. I can add a response :)

avivace (approved):

invenio_cli/helpers/docker_helper.py
ppanero: Fair, still better safe than raise on runtime 😅

ppanero (approved):

PR #330, opened 2022-10-25
global: add support for translation commands

ppanero:
- closes https://github.com/inveniosoftware/cookiecutter-invenio-rdm/issues/241

In the following image you can see:

- On the top left, extracted messages both from `./site` and `invenio.cfg`
- On the bottom left the part of the `container start ...` that corresponds to the catalog compilation
- On the right, InvenioRDM running on containerized version (alsto tested local and it's the same) with a translated string being shown.

Screenshot 2022-10-26 at 16 58 52

**Pending**
- If the implementation is approved/merged. We should document it, since users still need to create catalogs for the desired language (that cannot be automated since we cannot know in advance which language/s the user wants).

---

@mb-wali This PR (along with one in cookiecutter) will enable users to add their own custom translations + translate strings form their cfg/custom code. Requesting your review since you are Translator master 🚀

### DIFF
--- a/invenio_cli/cli/translations.py
+++ b/invenio_cli/cli/translations.py
@@ -1,0 +1,93 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2022 CERN.
+#
+# Invenio-Cli is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""Invenio module to ease the creation and management of applications."""
+
+from pathlib import Path
+
+import click
+
+from ..commands import TranslationsCommands
+from .utils import pass_cli_config, run_steps
+
+
+@click.group()
+def translations():
+    """Commands for translations management."""
+
+
+@translations.command()
+@click.option(
+    "--babel-ini",
+    "-b",
+    default="translations/babel.ini",
+    help="Relative path to babel.ini (including filename).",
+)
+@pass_cli_config
+def extract(cli_config, babel_ini):
+    """Extract messages for i18n support (translations)."""
+    click.secho("Extracting messages...", fg="green")
+    steps = TranslationsCommands.extract(
+        msgid_bugs_address=cli_config.get_author_email(),
+        copyright_holder=cli_config.get_author_name(),
+        babel_file=cli_config.get_project_dir() / Path("translations/babel.ini"),
+        output_file=cli_config.get_project_dir() / Path("translations/messages.pot"),
+        input_dirs=cli_config.get_project_dir(),
+    )
+    on_fail = "Failed to extract messages."
+    on_success = "Messages extracted successfully."
+
+    run_steps(steps, on_fail, on_success)
+
+
+@translations.command()
+@click.option("--locale", "-l", required=True, help="Locale to initialize.")
+@pass_cli_config
+def init(cli_config, locale):
+    """Initialized message catalog for a given locale."""
+    click.secho("Initializing messages catalog...", fg="green")
+    steps = TranslationsCommands.init(
+        output_dir=cli_config.get_project_dir() / Path("translations/"),
+        input_file=cli_config.get_project_dir() / Path("translations/messages.pot"),
+        locale=locale,
+    )
+    on_fail = f"Failed to initialize message catalog for {locale}."
+    on_success = f"Message catalog for {locale} initialized successfully."
+
+    run_steps(steps, on_fail, on_success)
+
+
+@translations.command()
+@pass_cli_config
+def update(cli_config):
+    """Update messages catalog."""
+    click.secho("Updating messages catalog...", fg="green")
+    steps = TranslationsCommands.update(
+        output_dir=cli_config.get_project_dir() / Path("translations/"),
+        input_file=cli_config.get_project_dir() / Path("translations/messages.pot"),
+    )
+    on_fail = f"Failed to update message catalog."
+    on_success = f"Message catalog updated successfully."
+
+    run_steps(steps, on_fail, on_success)
+
+
+@translations.command()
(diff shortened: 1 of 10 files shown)


invenio_cli/commands/translations.py
```diff
@@ -0,0 +1,54 @@
...
+from pathlib import Path
+
+from ..commands import Commands
+from .steps import CommandStep
+
+
+class TranslationsCommands(Commands):
+    """Translations CLI commands."""
+
+    CMD_PREFIX = ["pipenv", "run"]
```
ppanero, line 19: This is a common pitfall across the whole library. It should potentially be moved up to the `Commands` class, or to the command class since `pipenv run...` is a common prefix for all `CommandSteps`.

invenio_cli/commands/translations.py
```diff
@@ -0,0 +1,54 @@
...
+    def extract(
+        self,
+        babel_file=None,
+        msgid_bugs_address=None,
+        copyright_holder=None,
+        add_comments=None,
+    ):
+        """Extract messages from source code and templates."""
+
+        msgid_bugs_address = msgid_bugs_address or self.cli_config.get_author_email()
```
ppanero, line 30: Should the commands receive just parameters? and have no knowledge about the `cli_config`? We are passing it across the whole library but feels wrong...

invenio_cli/commands/translations.py
```diff
@@ -0,0 +1,120 @@
...
+        return [
+            CommandStep(
+                cmd=cmd,
+                env={"PIPENV_VERBOSITY": "-1"},
+                message=f"Initializing message catalog for {locale}...",
+            )
+        ]
+
+
+    def update(self):
```
ppanero, line 79: init and update are very similar, consolidate?

invenio_cli/commands/translations.py
```diff
@@ -0,0 +1,134 @@
...
+                message=f"Compiling message catalog...",
+            ),
+            FunctionStep(
+                func=self.symlink_translations,
+                args={"target": translation_folder},
+                message="Symlinking translations...",
+            ),
+        ]
+
+    def symlink_translations(self, target):
```
ppanero, line 129: This function is duplicated code form some other cmd class (install, and maybe more)

invenio_cli/cli/translations.py
```diff
@@ -0,0 +1,91 @@
...
+
+@translations.command()
+@click.option(
+    "--babel-ini", "-b", default="translations/babel.ini", help="Relative path to babel.ini (including filename)."
+)
+@pass_cli_config
+def extract(cli_config, babel_ini):
+    """Extract messages for i18n support (translations)."""
+    click.secho("Extracting messages...", fg="green")
+    steps = TranslationsCommands.extract(
```
ppanero, line 31: Pointing out issue: some _command_ classes are called as instance some as class. In some cases (e.g. services) the docker helper is needed (all good here). However, from the base class these classes inherit the need for a `cli_config`. Should this stay at `cli/` package level? and be passed as parameters if needed?

invenio_cli/commands/translations.py
```diff
@@ -0,0 +1,116 @@
...
+                env={"PIPENV_VERBOSITY": "-1"},
+                message=f"Initializing message catalog for {locale}...",
+            )
+        ]
+
+    @classmethod
+    def update(cls, output_dir, input_file):
+        """Update the message catalog."""
+
+        cmd = cls.CMD_PREFIX + [
```
ppanero, line 77: Seems to be almost equal to `init`. However, I do not find a semantically meaningful way of just sharing the cmd. It would make things ad-hoc... so duplication is known evil here.

invenio_cli/commands/containers.py
```diff
@@ -189,6 +216,29 @@ def static_pages(self, project_shortname):
...
 
+    def translations(self, project_shortname):
+        """Steps to compile translations for the instance."""
+        commands = TranslationsCommands(
+            project_path=self.cli_config.get_project_dir(),
+            instance_path=self.cli_config.get_instance_path(),
+        )
+        cmd = commands.compile(
+            # instance path inside the container
+            directory="/opt/invenio/var/instance/translations",
```
ppanero, line 227: This whole function feels like an ugly hack to account for a containers environment. The instance path should not change? Can we maybe just pass as string `{INVENIO_INSTANCE_PATH}` and assume that is set from the docker image? (still not super beautiful...)

invenio_cli/commands/containers.py
```diff
@@ -142,6 +143,32 @@ def _setup(self, project_shortname="/opt/var/instance/"):
             ),
         ]
 
+        if rdm_version()[0] >= 10:
+            steps.extend(
```
zzacharo, line 147: Thanks for fixing that!

zzacharo (approved):

invenio_cli/commands/containers.py
ppanero: @zzacharo Unfortunately no. The `get_instance_path` points to the local one stored in `.invenio.private` (e.g. `/Users/ppanero/.virtualenvs/instance-translations/var/instance`). In this case we need the instance path inside the container, namely https://github.com/inveniosoftware/docker-invenio/blob/master/almalinux/Dockerfile#L64.

zzacharo: Can we maybe write this information in .invenio.private so it can be used for the containerized version? Still hardcoded but at least users can tweak that...wdyt?

ppanero: Discussed in real life: going with using the interpolation of `${INVENIO_INSTANCE_PATH}`

zzacharo (approved):

PR #345, opened 2023-03-07
Services: add 'declare queues' step to setup

max-moser:
This is required in order for the statistics integration to work properly, because it relies on messaging via a message queue.
Needs some testing.

### DIFF
--- a/invenio_cli/commands/services.py
+++ b/invenio_cli/commands/services.py
@@ -220,6 +220,9 @@ class ServicesCommands(Commands):
             steps.extend(self.rdm_fixtures())
             steps.extend(self.translations())
 
+        if rdm_version()[0] >= 12:
+            steps.extend(self.declare_queues())
+
         steps.append(
             FunctionStep(
                 func=self.cli_config.update_services_setup,
@@ -240,6 +243,12 @@ class ServicesCommands(Commands):
             )
         ]
 
+        return steps
+
+    def declare_queues(self):
+        """Steps to declare the MQ queues required for statistics, etc."""
+        command = ["pipenv", "run", "invenio", "queues", "declare"]
+        steps = [CommandStep(cmd=command, message="Declaring queues...")]
         return steps
 
     def fixtures(self):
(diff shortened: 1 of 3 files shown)


max-moser:
Seems to work well enough, the queues configured for statistics are indeed declared after `invenio services setup` and the statistics are working.

kpsherva (approved):

kpsherva:
@slint @ntarocco @zzacharo should we make this step optional, called on demand or we always want to declare ?

ntarocco:
I think this should be always there, mandatory step. Why optional?

max-moser:
One thing that I just noticed (b/c I've just tested it with the master-branch InvenioRDM):
This command will probably fail with older versions of InvenioRDM, because Invenio-Queues is only added as dependency in v12 :eyes:

invenio_cli/commands/services.py
```diff
@@ -296,6 +302,7 @@ def setup(self, force, demo_data=True, stop=False, services=True):
 
         steps.extend(self._setup())
         steps.extend(self.fixtures())
+        steps.extend(self.declare_queues())
```
ppanero, line 305: As Max pointed out `invenio-queues` only comes after v12. Therefore it needs an `rdm_version >= 12` if-condition. I am wondering if this could be moved inside `_setup` (since it should run before fixtures in case we for example want demo stats).  Then it could be moved below this `if`:

https://github.com/inveniosoftware/invenio-cli/blob/f165c58ceafa15c1518ebbe528039432b1570afe/invenio_cli/commands/services.py#L219

invenio_cli/commands/services.py
```diff
@@ -255,6 +255,12 @@ def fixtures(self):
 
         return steps
 
+    def declare_queues(self):
```
ppanero, line 258: major: this will work for a local setup, it also needs to be included in the containerized version since they are `FunctionStep`.

invenio_cli/commands/services.py
ntarocco: Very good point! 👍🏼

kpsherva (approved):

ppanero (approved):

Issue #348, opened 2023-05-10
docker-compose reaches EOL in June 


wgresshoff:
## Is your feature request related to a problem? Please describe.
docker-compose v1 reaches EOL end of june 2023. Additionally the syntax changes: instead of `docker-compose` new versions add `compose` as subcommand to docker, so it is `docker compose` instead.

## Describe the solution you'd like
Support `docker-compose` if available and the newe syntax `docker compose`.

## Describe alternatives you've considered
...

## Additional context
Official announcement: https://docs.docker.com/compose/reference/

mfenner:
I have been using docker compose V2 and the new syntax `docker compose up` for some time now and have not encountered any problems. Example at https://github.com/front-matter/docker-invenio-rdm

mfenner:
@karkraeg I used invenio-cli to build the docker images used in my Docker Compose setup.

karkraeg:
@fenekku @ppanero I modified my Fork to use `docker compose` instead of `docker-compose` v1: https://github.com/inveniosoftware/invenio-cli/commit/8d77c22dd207ad385e50df29a44e8327526b0336

Would this change be of interest as a Pull Request?

`docker-compose`'s last release was more than two years ago: https://docs.docker.com/compose/migrate/#will-i-still-be-able-to-use-compose-v1-if-i-really-want-to

This would alter the installation dependencies for invenioRDM.

fenekku:
@karkraeg Yes that would be great! You can make a PR on this repo and associate it with the "PR Community" project (I will be on the lookout for it anyway if there are logistical issues). 

What I am thinking is that we could either have a transition period where both are supported  (recent versions of docker install a `docker-compose` pass-through anyway) or make a clean break especially for upcoming v12. 

Clean break is probably easier and installing most recent version of docker is possible in any OS (we were initially a holdout because we didn't think our setup could support it, but it can!).

mfenner:
Any updates on this?

karkraeg:
> Any updates on this?

Hi, I'm currently on vacation, I will look into whats missing to merge the PR on the 24th of July or perhaps next week, but can't promise

mfenner:
Thanks for the update. Docker-compose V1 has reached EOL in June, so there is some urgency to resolve this.

PR #352, opened 2023-06-06
Switch to `docker compose` v2 

karkraeg:
* closes https://github.com/inveniosoftware/invenio-cli/issues/348

### Description

As described in https://github.com/inveniosoftware/invenio-cli/issues/348 `invenio-cli` internally uses the deprecated version 1x of `docker-compose` to orchestrate containers. This forces users to install the old version alongside the newer version of `docker compose`. This pull request changes the underlying subcommands to use the new syntax.

### Checklist

- [X] I'm aware of the [code of conduct](https://inveniordm.docs.cern.ch/contribute/code-of-conduct/).
- [X] I've identified the [copyright holder(s)](https://inveniordm.docs.cern.ch/contribute/copyright-policy/) and updated copyright headers for touched files (>15 lines contributions).
- [X] I've NOT included third-party code (copy/pasted source code or new dependencies).

**Reminder**

### DIFF
--- a/invenio_cli/commands/requirements.py
+++ b/invenio_cli/commands/requirements.py
@@ -196,8 +196,8 @@ class RequirementsCommands(object):
     def check_dev(cls):
         """Steps to check the development pre-requisites."""
         if rdm_version()[0] >= 11:
-            node_version = 18
-            npm_version = 9
+            node_version = 16
+            npm_version = 7
         else:
             # for backwards compatibility with v9 (LTS)
             node_version = 14
--- a/invenio_cli/commands/requirements.py
+++ b/invenio_cli/commands/requirements.py
@@ -2,6 +2,7 @@
 #
 # Copyright (C) 2020-2021 CERN.
 # Copyright (C) 2021 TU Wien.
+# Copyright (C) 2023 ULB Münster.
 #
 # Invenio-Cli is free software; you can redistribute it and/or modify it
 # under the terms of the MIT License; see LICENSE file for more details.
@@ -113,7 +114,7 @@ class RequirementsCommands(object):
         # Output comes in the form of
         # 'Docker version 19.03.13, build 4484c46d9d\n'
         try:
-            result = run_cmd(["docker", "--version"])
+            result = run_cmd(["docker", "version"])
             version = cls._version_from_string(result.output.strip())
             return cls._check_version("Docker", version, major, minor, patch, exact)
         except Exception as err:
@@ -123,9 +124,9 @@ class RequirementsCommands(object):
     def check_docker_compose_version(cls, major, minor=-1, patch=-1, exact=False):
         """Check the docker compose version."""
         # Output comes in the form of
-        # 'docker-compose version 1.27.4, build 4484c46d9d\n'
+        # 'Docker Compose version v2.17.3\n'
         try:
-            result = run_cmd(["docker-compose", "--version"])
+            result = run_cmd(["docker", "compose", "version"])
             version = cls._version_from_string(result.output.strip())
             return cls._check_version(
                 "Docker Compose", version, major, minor, patch, exact
@@ -193,8 +194,8 @@ class RequirementsCommands(object):
     def check_dev(cls):
         """Steps to check the development pre-requisites."""
         if rdm_version()[0] >= 11:
-            node_version = 16
-            npm_version = 7
+            node_version = 18
+            npm_version = 9
         else:
             # for backwards compatibility with v9 (LTS)
             node_version = 14
--- a/invenio_cli/commands/requirements.py
+++ b/invenio_cli/commands/requirements.py
@@ -13,10 +13,10 @@ import re
 import sys
 from os import listdir
 
+from ..helpers.docker_helper import DockerHelper
 from ..helpers.process import ProcessResponse, run_cmd, run_interactive
 from ..helpers.rdm import rdm_version
 from .steps import FunctionStep
-from ..helpers.docker_helper import DockerHelper
 
 
 class RequirementsCommands(object):
--- a/invenio_cli/commands/requirements.py
+++ b/invenio_cli/commands/requirements.py
@@ -9,6 +9,7 @@
 
 """Invenio module to ease the creation and management of applications."""
 
+import json
 import re
 import sys
 from os import listdir
@@ -112,11 +113,11 @@ class RequirementsCommands(object):
     @classmethod
     def check_docker_version(cls, major, minor=-1, patch=-1, exact=False):
         """Check the docker version."""
-        # Output comes in the form of
-        # 'Docker version 19.03.13, build 4484c46d9d\n'
+        # Use JSON Formatted output and parse it
         try:
-            result = run_cmd(["docker", "version"])
-            version = cls._version_from_string(result.output.strip())
+            result = run_cmd(["docker", "version", "--format", "json"])
+            result_json = json.loads(result.output.strip())
+            version = cls._version_from_string(result_json["Client"]["Version"])
             return cls._check_version("Docker", version, major, minor, patch, exact)
         except Exception as err:
             return ProcessResponse(error=f"Docker not found. Got {err}.", status_code=1)
--- a/invenio_cli/commands/requirements.py
+++ b/invenio_cli/commands/requirements.py
@@ -16,6 +16,7 @@ from os import listdir
 from ..helpers.process import ProcessResponse, run_cmd, run_interactive
 from ..helpers.rdm import rdm_version
 from .steps import FunctionStep
+from ..helpers.docker_helper import DockerHelper
 
 
 class RequirementsCommands(object):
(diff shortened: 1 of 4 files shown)


karkraeg:
Hi @fenekku I can't associate it with the "PR Community" project

fenekku (approved):
Thanks for the contribution! The npm/node version changes is the only thing that should really be put aside for now.

invenio_cli/helpers/docker_helper.py
```diff
@@ -34,7 +35,7 @@ def _normalize_name(self, project_shortname):
         Docker-Compose introduced support for dash and underscore in
         version 1.21.0.
         """
-        dc_version_string = run_cmd(["docker-compose", "--version"])
+        dc_version_string = run_cmd(["docker", "compose", "version"])
```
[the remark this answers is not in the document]
karkraeg: I'll look into this, will report later!

invenio_cli/commands/requirements.py
```diff
@@ -193,8 +194,8 @@ def check_pipenv_installed(cls):
     def check_dev(cls):
         """Steps to check the development pre-requisites."""
         if rdm_version()[0] >= 11:
-            node_version = 16
-            npm_version = 7
+            node_version = 18
+            npm_version = 9
```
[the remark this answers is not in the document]
fenekku: I don't know of a way to keep those out of the PR from our end. Yeah, make a new commit removing them and we can squash and merge when integrating the work.

invenio_cli/commands/requirements.py
```diff
@@ -125,8 +126,9 @@ def check_docker_compose_version(cls, major, minor=-1, patch=-1, exact=False):
         """Check the docker compose version."""
         # Output comes in the form of
         # 'Docker Compose version v2.17.3\n'
+        docker_helper = DockerHelper("", local=False)
```
karkraeg, line 129: I guess this is not how it's supposed to be, but I can't call `cli_config.get_project_shortname()` as in https://github.com/inveniosoftware/invenio-cli/blob/5f89470b8015b716a4b63f3163214595ff7f74ee/invenio_cli/commands/services.py#L30 in this context.

invenio_cli/commands/requirements.py
```diff
@@ -113,7 +114,7 @@ def check_docker_version(cls, major, minor=-1, patch=-1, exact=False):
         # Output comes in the form of
         # 'Docker version 19.03.13, build 4484c46d9d\n'
         try:
-            result = run_cmd(["docker", "--version"])
+            result = run_cmd(["docker", "version"])
```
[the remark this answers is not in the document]
slint: We list in the [system requirements docs](https://inveniordm.docs.cern.ch/install/requirements/) `Docker 20.10.10+` since [we migrated the Docker images](https://github.com/inveniosoftware/docs-invenio-rdm/pull/513). 

As you said, there might be dangers with just upgrading Docker and running commands on existing containers/setups, but I think we never advertised `docker-compose` as a production setup so it's only going to be an issue during development.

I would say this is a minor bump, given how close we are to the June EOL of Docker v1.

invenio_cli/helpers/docker_helper.py
```diff
@@ -24,6 +25,7 @@ class DockerHelper(object):
     def __init__(self, project_shortname, local=True, log_config=None):
         """Constructor."""
         super().__init__()
+        self.docker_compose = ["docker", "compose"]
```
slint, line 28: comment: at this point I'm even sceptical of if we need the class attribute `self.docker_compose`, or if we can just replace everywhere `docker-compose` -> `docker compose`.

Not blocking for merging this PR, but something to possibly shelve and explore later as a task to simplify the codebase.

jennur (approved):
Looks good to us, but agree with Alex that we should be consistent with hardcoding `["docker", "compose"]`or using the variable `docker_compose`.

invenio_cli/commands/requirements.py
```diff
@@ -113,7 +115,7 @@ def check_docker_version(cls, major, minor=-1, patch=-1, exact=False):
...
```
ppanero, line 118: update comment in line 19

ppanero (approved):
LGTM, I think the corresponding docs should also be updated when/before this is merged.

invenio_cli/commands/requirements.py
karkraeg: `docker version` outputs the version like this:

```
Client: Docker Engine - Community
 Version:           24.0.2
 API version:       1.43
 ...
```

as opposed to `docker -version`: `Docker version 24.0.2, build cb74dfc`

The classmethod `_version_from_string` works because it selects the first regex match, but I think it would actually be better to keep the `--version` query. 

Or which comment are you referring to?

slint: A more "reliable" alternative would be to use `docker version --format json` which returns JSON with the version string under the `Client.Version` key

invenio_cli/commands/requirements.py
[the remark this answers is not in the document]
zzacharo: @slint @fenekku Given the impact of this change, even if I agree that is not breaking change for the public API per see, I am leaning towards a major.... Even if it doesn't fall in this category https://semver.org/#what-if-i-inadvertently-alter-the-public-api-in-a-way-that-is-not-compliant-with-the-version-number-change-ie-the-code-incorrectly-introduces-a-major-breaking-change-in-a-patch-release I like the answer given there
```
Remember, Semantic Versioning is all about conveying meaning by how the version number changes. If these changes are important to your users, use the version number to inform them.
```
WDYT?

invenio_cli/commands/requirements.py
karkraeg: This would mean that we would have to chamnge [`check_docker_version`](https://github.com/inveniosoftware/invenio-cli/blob/master/invenio_cli/commands/requirements.py#L111) to something like this, should I add this to this PR?

```python
@classmethod
def check_docker_version(cls, major, minor=-1, patch=-1, exact=False):
    """Check the docker version."""
    # Use JOSN Formatted output and parse it
    try:
        result = run_cmd(["docker", "version", "--format", "json"])
        result_json = json.loads(result.output.strip())
        version = cls._version_from_string(result_json["Client"]["Version"])
        return cls._check_version("Docker", version, major, minor, patch, exact)
    except Exception as err:
        return ProcessResponse(error=f"Docker not found. Got {err}.", status_code=1)         
```

invenio_cli/commands/requirements.py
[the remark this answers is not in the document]
fenekku: Yeah it's not super clear cut and semantic versioning creates or struggles with weird marketing expectations (From that link: "Semantic Versioning is all about conveying meaning by how the version number changes. If these changes are important to your users, use the version number to inform them." -- that's a lot of leeway :sweat: ).

I lean for a minor bump per the "it's the dependency that changed and not the project's API" argument, as well as @slint's note on the pre-existing requirement change that didn't lead to a major bump. I think of docker-compose as a regualr python dependency/python version now and how there are typically no major version bump in a project as an older python version/package version is deprecated in favor of a new one.

zzacharo (approved):

invenio_cli/commands/requirements.py
```diff
@@ -123,9 +125,10 @@ def check_docker_version(cls, major, minor=-1, patch=-1, exact=False):
     def check_docker_compose_version(cls, major, minor=-1, patch=-1, exact=False):
         """Check the docker compose version."""
         # Output comes in the form of
-        # 'docker-compose version 1.27.4, build 4484c46d9d\n'
+        # 'Docker Compose version v2.17.3\n'
+        docker_helper = DockerHelper("", local=False)
         try:
-            result = run_cmd(["docker-compose", "--version"])
+            result = run_cmd(docker_helper.docker_compose + ["version"])
```
zzacharo, line 131:
```suggestion
            result = run_cmd("docker", "compose", "version"])
```
wouldn't that work?

invenio_cli/commands/requirements.py
slint: LGTM and I think it brings us a bit closer to having a "stable" way of parsing the output, and if the option is there by Docker, I imagine they plan to keep it around.

invenio_cli/commands/requirements.py
slint: Yes, but as discussed IRL, let's shelve for now and merge since it might be a bigger task to remove the helper usage across the module.

slint (approved):
LGTM, one comment only to shelve and address later.

karkraeg:
Hi, I updated my fork with your latest changes. Is there anything blocking the merging that I could help with? Tbh I somehow lost track.

fenekku:
Everyone approved, tests pass and it's been out very long: I am merging. Thank you @karkraeg for all the effort, we appreciate it!

PR #367, opened 2024-05-23
Split `install` and `run` commands

slint:
* Split `install` command to `install {symlink,python,assets,all}`.
* Split `run` command to `run {web,worker,all}`.
* Explicitly makes the Celery worker listen to the "default" and "low" queues.


utnapischtim:
superseded by #383

PR #371, opened 2024-07-02
wip: transition to uv

utnapischtim:
* this gives a huge performance improvement on installing and setting up
  an environment.

NOTE:
The problem is that the commands `outdated_packages`, `update_packages`, `update_package_new_version` are at the moment not that easy to port to `uv`.  this is a argument against moving to `uv` but i hope that the maintainer of `uv` will improve `uv` in that direction. in the meantime we could figure out how it could be implemented with a few intermediate steps, maybe they are then at least as fast as the current approach.


utnapischtim:
superseded by #383

PR #372, opened 2024-07-02
make container alpine ready

utnapischtim:
- service: add configurable host and port for search
- services: change from bash to sh
- chore: remove unused statements
- services: rely on self.local for stop and destroy

### DIFF
--- a/invenio_cli/commands/services_health.py
+++ b/invenio_cli/commands/services_health.py
@@ -29,8 +29,6 @@ class ServicesHealthCommands(object):
     @classmethod
     def search_healthcheck(cls, *args, **kwargs):
         """Open/Elasticsearch healthcheck."""
-        verbose = kwargs["verbose"]
-
         host = kwargs["search_host"]
         port = kwargs["search_port"]
         return run_cmd(
@@ -41,7 +39,6 @@ class ServicesHealthCommands(object):
     def postgresql_healthcheck(cls, *args, **kwargs):
         """Postgresql healthcheck."""
         filepath = kwargs["filepath"]
-        verbose = kwargs["verbose"]
 
         return run_cmd(
             [
@@ -62,7 +59,6 @@ class ServicesHealthCommands(object):
     def mysql_healthcheck(cls, *args, **kwargs):
         """Mysql healthcheck."""
         filepath = kwargs["filepath"]
-        verbose = kwargs["verbose"]
         password = kwargs["project_shortname"]
 
         return run_cmd(
@@ -84,7 +80,6 @@ class ServicesHealthCommands(object):
     def redis_healthcheck(cls, *args, **kwargs):
         """Redis healthcheck."""
         filepath = kwargs["filepath"]
-        verbose = kwargs["verbose"]
 
         return run_cmd(
             [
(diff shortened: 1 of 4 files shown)


invenio_cli/helpers/docker_helper.py
```diff
@@ -98,14 +99,18 @@ def start_containers(self, app_only=False):
 
     def stop_containers(self):
         """Stop currently running containers."""
-        command = self.docker_compose + ["--file", "docker-compose.full.yml", "stop"]
+        command = self.docker_compose + [
```
fenekku, line 102: :heart: we have had a custom script to get around this for the longest time

invenio_cli/commands/services_health.py
```diff
@@ -49,7 +49,7 @@ def postgresql_healthcheck(cls, *args, **kwargs):
                 "exec",
                 "-T",
                 "db",
-                "bash",
+                "sh",
```
fenekku, line 52: The idea here being that one may have modified their docker-compose.yml file to use an alpine based image for their postgresql or redis service... With `sh` being on both images this works.

invenio_cli/commands/services_health.py
```diff
@@ -28,17 +29,16 @@ class ServicesHealthCommands(object):
     @classmethod
     def search_healthcheck(cls, *args, **kwargs):
         """Open/Elasticsearch healthcheck."""
-        verbose = kwargs["verbose"]
-
+        host = kwargs["search_host"]
+        port = kwargs["search_port"]
         return run_cmd(
-            ["curl", "-f", "localhost:9200/_cluster/health?wait_for_status=yellow"]
+            ["curl", "-f", f"{host}:{port}/_cluster/health?wait_for_status=yellow"]
```
fenekku, line 35: Was this part of allowing a local instance and the tests to run at the same time? I guess it could be tolerant to multiple document engines running on different ports e.g., work on different projects and this makes it easier to switch between them. 

The host part I am less convinced by, but we are letting people pass a host and port for the server already, so c'est-la-vie. In general, I think we should steer people away from using invenio-cli to run elaborate multi-machine setup and a good way is to not let them. Power users can usually figure it out anyway (have the skill to create workaround scripts). For newcomers it's another way to get confused. But here it's done tastefully and there is prior art, so that's fine.

fenekku (approved):
Thanks. This all works and stays within the bounds. By that I mean, that I wouldn't be supportive of having to develop invenio-cli in such a way that outlandishly modified docker-compose.yml files have to be supported. Any more than that, and I think we shouldn't do it or really rethink what invenio-cli is what abstractions it should use to frame its functionality.

(I will let this sit a little more and then we can merge if all good)

invenio_cli/commands/services_health.py
utnapischtim: yes this changes makes it possible to run multiple instances at the same time. the thing is i couldn't make it work without that change. using only docker-compose files is not enough because the healthcheck in invenio-cli blocks the `invenio-cli run` command.

but those `kwargs["search_host"]` values have default values so the current behavior still works!

invenio_cli/commands/services_health.py
utnapischtim: yes. `bash` is to heavy for `alpine` ;)

fenekku:
Ok, wrapping my day and this is good. Merging. Thanks @utnapischtim !

PR #373, opened 2024-07-04
services: Add setup steps for ILS

sakshamarora1:


invenio_cli/commands/services.py
```diff
@@ -193,43 +194,62 @@ def _setup(self):
             ),
         ]
 
-        if rdm_version()[0] >= 10:
+        rdm_version_value = rdm_version()
+        if rdm_version_value:
```
kpsherva, line 198: I would suggest to split the conditional statement into two separate files (one per app), because it is becoming sizeable considering different rdm version and adding ILS, but I would like to ask fir input also from other reviewers

ntarocco: Nothing against this, I just don't know how hard or easy it is :)

ntarocco (approved):

PR #375, opened 2024-08-01
v12 release

fenekku:
- **dependencies: update for invenio-app-rdm v12**
- **release: v1.5.0**

### DIFF
--- a/invenio_cli/commands/requirements.py
+++ b/invenio_cli/commands/requirements.py
@@ -196,7 +196,10 @@ class RequirementsCommands(object):
     @classmethod
     def check_dev(cls):
         """Steps to check the development pre-requisites."""
-        if rdm_version()[0] >= 11:
+        if rdm_version()[0] >= 12:
+            node_version = 18
+            npm_version = 10
+        elif rdm_version()[0] >= 11:
             node_version = 16
             npm_version = 7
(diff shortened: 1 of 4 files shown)


invenio_cli/commands/requirements.py
```diff
@@ -196,7 +196,10 @@ def check_pipenv_installed(cls):
     @classmethod
     def check_dev(cls):
         """Steps to check the development pre-requisites."""
-        if rdm_version()[0] >= 11:
+        if rdm_version()[0] >= 12:
```
fenekku, line 199: Honestly we should revisit this afterwards since this has a cost to maintain/gets out of sync.

utnapischtim: i thought the same. Since we thought about v12 is the next LTS we could skip the v9 support?

utnapischtim (approved):

invenio_cli/commands/requirements.py
fenekku: We support v9 until Feb 2025, so we can remove it then.

fenekku:
Once merged and released (I should be able to), I'll merge the docs too.

slint (approved):
:shipit:

PR #383, opened 2025-02-19
uv rspack pnpm web worker ports

utnapischtim:
this supersede #380  

# steps to how to use it

1.) check that `uv` and `pnpm` are installed
2.) create a new my-site directory with the regular content as usual OR go into your current development my-site directory
2.a) deactivate the active environment
3.) create with uv a virtual environment with `uv venv --prompt uv-env && source .venv/bin/activate`
4.) install invenio-cli with `uv pip install "git+https://github.com/utnapischtim/invenio-cli@uv-rspack-pnpm-web-worker-ports"`
4.a) `uv pip install "git+https://github.com/utnapischtim/pynpm@add-pnpm"`
5.) copy following text into the `.invenio` file into the cli section
```
python_packages_manager = uv
javascript_packages_manager = pnpm
assets_builder = rspack
```
5.a) please copy following to `invenio.cfg`
```
JAVASCRIPT_PACKAGES_MANAGER = "pnpm"
ASSETS_BUILDER = "rspack"
```

6.) copy following text into the `pyproject.toml` file. the file should be created beside the `invenio.cfg` file
```
[project]
name = "InvenioRDM"
requires-python = ">= 3.12"
dynamic = ["version"]

dependencies = [
  "invenio-app-rdm[opensearch2]~=13.0.0b2.dev2",
  "uwsgi>=2.0",
  "uwsgitop>=0.11",
  "uwsgi-tools>=1.1.1",

  # rspack
  "flask-webpackext",
  "invenio-assets",
  "pywebpack",
  "pynpm",
]

[tool.setuptools]
py-modules = [] # necessary to make the packages with setup.py usable with uv

# overrides packages from "dependencies"
[tool.uv.sources]
flask-webpackext = { git = "https://github.com/utnapischtim/flask-webpackext", branch = "make-ready-for-rspack" }
invenio-assets = { git = "https://github.com/slint/invenio-assets", branch = "rspack" }
pywebpack = { git = "https://github.com/utnapischtim/pywebpack", branch = "add-scripts" }
pynpm = { git = "https://github.com/utnapischtim/pynpm", branch = "add-pnpm" }
```
7.) run invenio-cli install to install

PR #384, opened 2025-02-28
Implement `uv` as an optional python package management backend

max-moser:
# Context
This is a variant of #383 with the `pnpm`/`rspack` stuff stripped out, and some squashed commits.
It only touches `uv`, since that's quite uncontroversial and works well enough already.

# How to set the Python package manager
If you have a `Pipfile` in your project, then `pipenv` will be used by default.
If you have `pyproject.toml` and no `Pipfile`, then `uv` will be used by default.
If your use case is not satisfied by this auto-magic, then you can explicitly configure the package manager to use in `.invenio`:
```ini
[cli]
flavour = RDM
python_package_manager = uv
```
Note the lack of quotes around the value `uv`.
Currently, only `uv` and `pipenv` are supported as values.

### DIFF
--- a/invenio_cli/commands/commands.py
+++ b/invenio_cli/commands/commands.py
@@ -32,7 +32,7 @@ class Commands(object):
     def pyshell(self, debug=False):
         """Start a Python shell."""
         pkg_man = self.cli_config.python_package_manager
-        with env(FLASK_DEBUG=str(debug)):
+        with env(FLASK_DEBUG="1" if debug else "0"):
             command = pkg_man.run_command("invenio", "shell")
             return run_interactive(command, env={"PIPENV_VERBOSITY": "-1"})
 
--- a/invenio_cli/commands/commands.py
+++ b/invenio_cli/commands/commands.py
@@ -7,6 +7,8 @@
 
 """Invenio module to ease the creation and management of applications."""
 
+
+from ..helpers.cli_config import CLIConfig
 from ..helpers.env import env
 from ..helpers.process import run_interactive
 from .steps import CommandStep
@@ -15,27 +17,23 @@ from .steps import CommandStep
 class Commands(object):
     """Abstraction over CLI commands that are either local or containerized."""
 
-    def __init__(self, cli_config):
+    def __init__(self, cli_config: CLIConfig):
         """Constructor.
 
         :param cli_config: :class:CLIConfig instance
         """
         self.cli_config = cli_config
 
-    @classmethod
-    def shell(cls):
+    def shell(self):
         """Start a shell in the virtual environment."""
-        command = [
-            "pipenv",
-            "shell",
-        ]
+        command = self.cli_config.python_package_manager.start_activated_subshell()
         return run_interactive(command, env={"PIPENV_VERBOSITY": "-1"})
 
-    @classmethod
-    def pyshell(cls, debug=False):
+    def pyshell(self, debug=False):
         """Start a Python shell."""
+        pkg_man = self.cli_config.python_package_manager
         with env(FLASK_DEBUG=str(debug)):
-            command = ["pipenv", "run", "invenio", "shell"]
+            command = pkg_man.run_command("invenio", "shell")
             return run_interactive(command, env={"PIPENV_VERBOSITY": "-1"})
 
     def destroy(self):
@@ -44,9 +42,10 @@ class Commands(object):
         NOTE: This function has no knowledge of the existence of services.
               Refer to services.py to destroy services' containers.
         """
+        command = self.cli_config.python_package_manager.remove_venv()
         steps = [
             CommandStep(
-                cmd=["pipenv", "--rm"],
+                cmd=command,
                 env={"PIPENV_VERBOSITY": "-1"},
                 message="Destroying virtual environment",
             )
--- a/invenio_cli/commands/local.py
+++ b/invenio_cli/commands/local.py
@@ -10,6 +10,7 @@
 
 import os
 import signal
+import sys
 from distutils.dir_util import copy_tree
 from os import environ
 from pathlib import Path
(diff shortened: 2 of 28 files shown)


invenio_cli/helpers/package_managers.py
```diff
@@ -0,0 +1,173 @@
...
+# Copyright (C) 2025 TU Wien.
+#
+# Invenio-Cli is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""Wrappers around various package managers to be used under the hood."""
+
+import os
+from abc import ABC
+from typing import List
```
max-moser, line 12: if you prefer, we can strip those type hints out again. i just added them for convenience during development, it did come in handy to avoid typos with my LSP.

invenio_cli/helpers/package_managers.py
```diff
@@ -0,0 +1,173 @@
...
+        return cmd
+
+    def lock_dependencies(self, prereleases, devtools):
+        """Update the lock file."""
+        cmd = [self.name, "lock"]
+        if prereleases:
+            cmd += ["--prerelease", "allow"]
+        return cmd
+
+    def remove_venv(self):
```
max-moser, line 162: i couldn't find `uv` commands for those; if you know more than me here, please let me know then we can update this.

invenio_cli/helpers/cli_config.py
```diff
@@ -61,6 +62,26 @@ def __init__(self, project_dir="./"):
...
+        )
+        if manager_name == Pipenv.name:
+            return Pipenv()
+        elif manager_name == UV.name:
+            return UV()
+
+        if (self.project_path / "Pipfile").is_file():
+            return Pipenv()
+        elif (self.project_path / "pyproject.toml").is_file():
+            return UV()
```
max-moser, line 79: personally i've found this quite convenient as it's not _necessary_ to add the config item to `.invenio` (although if present, that one takes precedence)

invenio_cli/commands/translations.py
max-moser, line 1: the translation commands were previously all static, but now we need to know about the configured python package manager

invenio_cli/commands/packages.py
```diff
@@ -10,24 +10,21 @@
...
-    def install_packages(packages, log_file=None):
-        """Steps to install Python packages.
-
-        It is a class method since it does not require any configuration.
-        """
-        prefix = ["pipenv", "run"]
-        cmd = prefix + ["pip", "install"]
-        for package in packages:
-            cmd.extend(["-e", package])
+    def __init__(self, cli_config: CLIConfig):
```
max-moser, line 21: type hint can be removed

invenio_cli/commands/commands.py
```diff
@@ -15,27 +17,23 @@
 class Commands(object):
     """Abstraction over CLI commands that are either local or containerized."""
 
-    def __init__(self, cli_config):
+    def __init__(self, cli_config: CLIConfig):
```
max-moser, line 20: type hint can be removed

invenio_cli/commands/upgrade.py
```diff
@@ -7,14 +7,18 @@
...
+from ..helpers.cli_config import CLIConfig
 from .steps import CommandStep
 
 
 class UpgradeCommands(object):
     """Local installation commands."""
 
-    @staticmethod
-    def upgrade(script_path):
+    def __init__(self, cli_config: CLIConfig):
```
max-moser, line 17: type hint can be removed

invenio_cli/commands/translations.py
```diff
@@ -7,26 +7,24 @@
...
 from .steps import CommandStep, FunctionStep
 
 
 class TranslationsCommands(Commands):
     """Translations CLI commands."""
 
-    CMD_PREFIX = ["pipenv", "run"]
-
-    def __init__(self, project_path, instance_path):
+    def __init__(self, cli_config: CLIConfig, project_path=None, instance_path=None):
```
max-moser, line 20: type hint can be removed

invenio_cli/commands/local.py
```diff
@@ -84,24 +84,23 @@ def update_statics_and_assets(self, force, debug=False, log_file=None):
...
         ops.append(self._statics)
-        ops.append(prefix + ["webpack", "build"])
+        ops.append(pkg_man.run_command("invenio", "webpack", "build"))
         # Keep the same messages for some of the operations for backward compatibility
         messages = {
             "build": "Building assets...",
             "install": "Installing JS dependencies...",
         }
 
-        with env(FLASK_DEUBG="true" if debug else "false"):
```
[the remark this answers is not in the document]
slint: Yeah, Flask checks for this in a bit of [a funny way](https://github.com/pallets/flask/blob/f61172b8dd3f962d33f25c50b2f5405e90ceffa5/src/flask/helpers.py#L32) to take into account all these cases. Agree, let's go with the `"0"`/`"1"` just to avoid the funny business with calling `str` on things.

invenio_cli/helpers/package_managers.py
slint: I guess we explicitly use `.venv` instead of getting it from config or similar? I'm not seeing that `pipenv --rm` is better in that manner, since it will just try to deduce based on its usual rules where the virtualenv is, e.g. `$HOME/Envs`, `.venv`, currently activated virtualenv (?), etc.

invenio_cli/helpers/cli_config.py
slint: Agree, and works well since we're explicitly checking for the file on the top-level (and not e.g. when someone randomly runs `invenio-cli ...` accidentally inside the `site` folder).

invenio_cli/helpers/package_managers.py
max-moser: i've looked for a way to make `uv` tell me where it's hiding its venv, but it wouldn't say.
apparently it only prints the location on creation.
until then i guess we'll just assume it's under the floor boards.

invenio_cli/helpers/cli_config.py
```diff
@@ -121,6 +142,14 @@ def get_search_host(self):
             "localhost",
         )
 
+    def get_web_port(self):
+        """Returns web port."""
+        return self.config[CLIConfig.COOKIECUTTER_SECTION].get("web_port", "5000")
```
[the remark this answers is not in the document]
max-moser: for the record, we just agreed to move this to `.invenio.private` instead as it's personal preference and not necessarily shared between developers.

slint (commented):
:shipit:

slint (approved):
:shipit:

PR #385, opened 2025-03-01
Enable PNPM (rspack is part of Invenio-Assets)

max-moser:
# Information (updated 2025-03-28)

This PR is the frontend cousin of https://github.com/inveniosoftware/invenio-cli/pull/384 (which similarly enabled `uv` for python).

## Enabling `pnpm`

Note: This feature requires `flask-webpackext` v2.1.0

Similar to `uv`, `pnpm` can be enabled in `.invenio` (which is the "public" configuration file that is intended to be shared between developers via version control).
Unlike `uv`, there are (currently [1]) no artifacts in the project directory that would indicate whether `npm` or `pnpm` should be used and thus the fallback is simply `npm` (i.e. `pnpm` has to be explicitly configured).

Configuration in `.invenio`:
```ini
[cli]
flavour = RDM
logfile = /logs/invenio-cli.log
javascript_package_manager = pnpm
```

[1] this may change once the "reusable frontend lock file" logic gets merged as:
* `package.lock` -> `npm`
* `pnpm-lock.yaml` -> `pnpm`

## Enabling `rspack`

Note: This feature requires `invenio-assets` v4.1.0 (or v3.2.0 for InvenioRDM v12)

`invenio.cfg`:
```py
# somewhere in the file...
WEBPACKEXT_PROJECT = "invenio_assets.webpack:rspack_project"
```

Like any other configuration item, it could alternatively be specified as an env var with prefix (`INVENIO_WEBPACKEXT_PROJECT="invenio_assets.webpack:rspack_project"`), but this is out of scope for here – iykyk.

## Why aren't `pnpm` and `rspack` both configured in `invenio.cfg`?

While it may seem strange that `rspack` uses a different configuration mechanism (`invenio.cfg`) than `pnpm` (`.invenio`), the reason here is simple:
1) The frontend build project was already previously configurable via a configuration item (old default: `WEBPACKEXT_PROJECT = "invenio_assets.webpack:project"`) that could for instance be configured via `invenio.cfg`.
2) The `pnpm` switch did **not** exist previously, and **both** `invenio-cli` as well as `invenio` (the application) have to agree on which JS package manager is used (and `invenio-cli` does not parse `invenio.cfg` itself).

## What's with `react-invenio-forms` and `pnpm`?

The [`link-dist` script in Invenio-y JS packages](https://github.com/inveniosoftware/react-invenio-forms/blob/master/package.json#L16) is geared towards `npm link` and thus obviously won't work nicely for `pnpm`.
(Also, requiring these nonstandard script entries means that only these Invenio-y JS packages are compatible.)
This should likely be reworked in these JS packages, but that's a bit too much effort for now.

For now, we've decided to simply write the `pnpm` logic to perform basically the same steps as `npm` would, except for the actual linking step (which only requires a single step with `pnpm`).
For those interested, I recommend searching the comments mentioning `prepostlink-dist` and others.

# More context (for historical reasons)

Second part of #383 – support for `pnpm` and `rspack`

This PR won't work as standalone, it still requires changes to core Invenio packages.
Either https://github.com/inveniosoftware/invenio-assets/pull/176
Or https://github.com/inveniosoftware/invenio-assets/pull/172, https://github.com/inveniosoftware/flask-webpackext/pull/26, https://github.com/inveniosoftware/pywebpack/pull/47
Depending on the choice, some tweaks need to be performed in this PR here.

~~Also, `npm` is still hard-coded in [React-Invenio-Forms](https://github.com/inveniosoftware/react-invenio-forms/blob/master/package.json#L16) which needs to be addressed.~~
Edit: This has been resolved, see above.

### DIFF
--- a/invenio_cli/commands/assets.py
+++ b/invenio_cli/commands/assets.py
@@ -45,20 +45,6 @@ class AssetsCommands(LocalCommands):
                 status_code=status_code,
             )
 
-    @staticmethod
-    def _run_script(module_pkg):
-        """Run script and return a ProcessResponse."""
-        status_code = module_pkg.run_script("link-dist")
-        if status_code == 0:
-            return ProcessResponse(
-                output="Module linked correctly to global", status_code=0
-            )
-        else:
-            return ProcessResponse(
-                error=f"Unable to link-dist. Got error code {status_code}",
-                status_code=status_code,
-            )
-
     def _npm_install_command(self, path, module_pkg):
         """Run command and return a ProcessResponse."""
         install_args = self.cli_config.javascript_package_manager.install_local_package(
@@ -85,28 +71,6 @@ class AssetsCommands(LocalCommands):
         else:
             return ProcessResponse(
                 error=f"Unable to build. Got error code {status_code}",
-                status_code=status_code,
-            )
-
-    @staticmethod
-    def _assets_link(assets_pkg, module_pkg):
-        try:
-            module_name = module_pkg.package_json["name"]
-        except FileNotFoundError as e:
-            return ProcessResponse(
-                error="No module found on the specified path. "
-                f"File not found {e.filename}",
-                status_code=1,
-            )
-
-        status_code = assets_pkg.link(module_name)
-        if status_code == 0:
-            return ProcessResponse(
-                output="Global module linked correctly to local folder", status_code=0
-            )
-        else:
-            return ProcessResponse(
-                error=f"Unable to link module. Got error code {status_code}",
                 status_code=status_code,
             )
 
@@ -142,17 +106,19 @@ class AssetsCommands(LocalCommands):
                 args={"module_pkg": module_pkg},
                 message="Building...",
             ),
-            FunctionStep(  # Create link to global folder
-                func=self._run_script,
-                args={"module_pkg": module_pkg},
-                message="Linking module to global dist...",
-            ),
-            FunctionStep(  # Link the global folder to the assets folder.
-                func=self._assets_link,
-                args={"assets_pkg": assets_pkg, "module_pkg": module_pkg},
-                message="Linking module to assets...",
-            ),
         ]
+
+        # The commands necessary for linking local JS packages vary by package manager
+        js_package_manager = self.cli_config.javascript_package_manager
+        link_steps = [
+            FunctionStep(
+                func=step.function,
+                args={"assets_pkg": assets_pkg, "module_pkg": module_pkg},
+                message=step.message,
+            )
+            for step in js_package_manager.package_linking_steps()
+        ]
+        steps.extend(link_steps)
 
         return steps
 
--- a/invenio_cli/commands/assets.py
+++ b/invenio_cli/commands/assets.py
@@ -7,11 +7,9 @@
 
 """Invenio module to ease the creation and management of applications."""
 
-import subprocess
 from pathlib import Path
 
 import click
-from pynpm import NPMPackage
 
 from ..helpers import env
 from ..helpers.process import ProcessResponse, run_interactive
@@ -26,17 +24,16 @@ class AssetsCommands(LocalCommands):
         """Constructor."""
         super().__init__(cli_config)
 
-    @staticmethod
-    def _module_pkg(path):
+    def _module_pkg(self, path):
         """NPM package for the given path."""
-        return NPMPackage(Path(path) / "package.json")
+        path = Path(path) / "package.json"
+        return self.cli_config.javascript_package_manager.create_pynpm_package(path)
 
     def _assets_pkg(self):
         """NPM package for the instance's webpack project."""
         return self._module_pkg(self.cli_config.get_instance_path() / "assets")
 
-    @staticmethod
-    def _watch_js_module(pkg):
+    def _watch_js_module(self, pkg):
         """Watch the JS module for changes."""
         click.secho("Starting watching module...", fg="green")
         status_code = pkg.run_script("watch")
@@ -62,10 +59,12 @@ class AssetsCommands(LocalCommands):
                 status_code=status_code,
             )
 
-    @staticmethod
-    def _npm_install_command(path):
+    def _npm_install_command(self, path, module_pkg):
         """Run command and return a ProcessResponse."""
-        status_code = subprocess.call(["npm", "install", "--prefix", path])
+        install_args = self.cli_config.javascript_package_manager.install_local_package(
+            path
+        )
+        status_code = module_pkg.install(" ".join(install_args))
         if status_code == 0:
             return ProcessResponse(
                 output="Dependent packages installed correctly", status_code=0
@@ -135,7 +134,7 @@ class AssetsCommands(LocalCommands):
         steps = [
             FunctionStep(  # Install dependent packages
                 func=self._npm_install_command,
-                args={"path": path},
+                args={"path": path, "module_pkg": module_pkg},
                 message="Installing dependent packages...",
             ),
             FunctionStep(  # Run build script
--- a/invenio_cli/commands/local.py
+++ b/invenio_cli/commands/local.py
@@ -84,16 +84,17 @@ class LocalCommands(Commands):
         Needed here (parent) because is used by Assets and Install commands.
         """
         # Commands
-        pkg_man = self.cli_config.python_package_manager
-        ops = [pkg_man.run_command("invenio", "collect", "--verbose")]
+        py_pkg_man = self.cli_config.python_package_manager
+        js_pkg_man = self.cli_config.javascript_package_manager
+        ops = [py_pkg_man.run_command("invenio", "collect", "--verbose")]
 
         if force:
-            ops.append(pkg_man.run_command("invenio", "webpack", "clean", "create"))
-            ops.append(pkg_man.run_command("invenio", "webpack", "install"))
+            ops.append(py_pkg_man.run_command("invenio", "webpack", "clean", "create"))
+            ops.append(py_pkg_man.run_command("invenio", "webpack", "install"))
         else:
-            ops.append(pkg_man.run_command("invenio", "webpack", "create"))
+            ops.append(py_pkg_man.run_command("invenio", "webpack", "create"))
         ops.append(self._statics)
-        ops.append(pkg_man.run_command("invenio", "webpack", "build"))
+        ops.append(py_pkg_man.run_command("invenio", "webpack", "build"))
         # Keep the same messages for some of the operations for backward compatibility
         messages = {
             "build": "Building assets...",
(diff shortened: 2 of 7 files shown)


invenio_cli/helpers/cli_config.py
```diff
@@ -82,6 +87,20 @@ def python_package_manager(self) -> PythonPackageManager:
...
+        """Get javascript packages manager."""
+        manager_name = self.config[CLIConfig.CLI_SECTION].get(
+            "javascript_package_manager"
+        )
+        if manager_name == NPM.name:
+            return NPM()
+        elif manager_name == PNPM.name:
+            return PNPM()
+
+        # fallback to NPM?
```
max-moser, line 101: `npm` would be the backwards-compatible default option

invenio_cli/commands/local.py
```diff
@@ -108,7 +109,9 @@ def update_statics_and_assets(self, force, debug=False, log_file=None):
                     if op[-1] in messages:
                         click.secho(messages[op[-1]], fg="green")
                     response = run_interactive(
-                        op, env={"PIPENV_VERBOSITY": "-1"}, log_file=log_file
+                        op,
+                        env={"PIPENV_VERBOSITY": "-1", **js_pkg_man.env_overrides()},
```
max-moser, line 113: here we tell `invenio` about the JS package manager that we want to use, via environment variables

invenio_cli/commands/assets.py
```diff
@@ -135,26 +163,40 @@ def link_js_module(self, path):
...
+                    args={"module_pkg": module_pkg},
+                    message="Linking module to global dist...",
+                ),
+                FunctionStep(  # Link the global folder to the assets folder.
+                    func=self._assets_link,
+                    args={"assets_pkg": assets_pkg, "module_pkg": module_pkg},
+                    message="Linking module to assets...",
+                ),
+            ]
+        else:
```
max-moser, line 189: In contrast to `npm`, `pnpm` supports linking via a single-step operation: https://pnpm.io/cli/link#pnpm-link-dir

invenio_cli/commands/local.py
```diff
@@ -84,16 +84,17 @@ def update_statics_and_assets(self, force, debug=False, log_file=None):
...
-            ops.append(pkg_man.run_command("invenio", "webpack", "clean", "create"))
-            ops.append(pkg_man.run_command("invenio", "webpack", "install"))
+            ops.append(py_pkg_man.run_command("invenio", "webpack", "clean", "create"))
+            ops.append(py_pkg_man.run_command("invenio", "webpack", "install"))
         else:
-            ops.append(pkg_man.run_command("invenio", "webpack", "create"))
+            ops.append(py_pkg_man.run_command("invenio", "webpack", "create"))
         ops.append(self._statics)
-        ops.append(pkg_man.run_command("invenio", "webpack", "build"))
+        ops.append(py_pkg_man.run_command("invenio", "webpack", "build"))
```
max-moser, line 97: there is still a good amount of optimization potential by calling the `invenio` command only once and thus saving all the redundant application initializations which eat up a lot of time.
one way to move forward is to implement a single larger command (e.g. in `invenio-app-rdm`) that performs all the steps currently implemented here, based on the provided CLI args.
@ntarocco has pointed out to me that something very much like that has already been done in [Invenio-App-ILS](https://github.com/inveniosoftware/invenio-app-ils/blob/master/invenio_app_ils/cli.py#L1567)

invenio_cli/helpers/cli_config.py
utnapischtim: yes, `npm` should be the default javascript package manager

invenio_cli/helpers/cli_config.py
```diff
@@ -82,6 +87,20 @@ def python_package_manager(self) -> PythonPackageManager:
                 "Could not determine the Python package manager, please configure it."
             )
 
+    @property
```
utnapischtim, line 90: should we use `functools.cached_property` here?

invenio_cli/helpers/cli_config.py
```diff
@@ -65,9 +72,7 @@ def __init__(self, project_dir="./"):
     @property
```
utnapischtim, line 72: should we use `functools.cached_property` here?

invenio_cli/helpers/package_managers.py
```diff
@@ -171,3 +174,57 @@ def start_activated_subshell(self) -> List[str]:
...
+        """Create an ``NPMPackage`` with the path to ``package.json``."""
+        return NPMPackage(package_json_path)
+
+    def install_local_package(self, path):
+        """Install the local JS package."""
+        return ["--prefix", str(path)]
+
+    def env_overrides(self):
+        """Provide environment overrides for building Invenio assets."""
+        return {"INVENIO_ASSETS_NPM_PKG_CLS": "pynpm:NPMPackage"}
```
utnapischtim, line 212: i think the name should not include NPM. i think something like `INVENIO_ASSETS_BUILDER_CLS` would be better

invenio_cli/commands/local.py
utnapischtim: an option would be to use a `rpc-server` implementation. a WIP would be [here](https://github.com/inveniosoftware/invenio-cli/pull/387) for `invenio-cli` combined with [PR](https://github.com/inveniosoftware/invenio-app/pull/99) from `invenio-app`

invenio_cli/commands/assets.py
```diff
@@ -135,26 +163,40 @@ def link_js_module(self, path):
...
-                message="Linking module to global dist...",
-            ),
-            FunctionStep(  # Link the global folder to the assets folder.
-                func=self._assets_link,
-                args={"assets_pkg": assets_pkg, "module_pkg": module_pkg},
-                message="Linking module to assets...",
-            ),
         ]
 
+        if self.cli_config.javascript_package_manager.name == "npm":
```
utnapischtim, line 176: would it be better to move that to the JavascriptPackageManager classes? it seems strange to have the classes and do such distinguishes here

invenio_cli/commands/assets.py
utnapischtim: i think i don't understand, but does this mean we don't have to change anything in the react-invenio-forms package? since if `pnpm` is configured, `_assets_link_single` is taking care of linking the package and it doesn't care as what is in the `react-invenio-forms/package.json` file?

invenio_cli/commands/assets.py
max-moser: this entire block is from the WIP commit which is only experimental so far.

invenio_cli/helpers/package_managers.py
max-moser: sure, we can change the naming although "builder" would be a bit unfortunate since the builder would be webpack/rspack, not npm/pnpm which are the package managers

utnapischtim: LOL, you are right, `BUILDER` is not the right word for that. i saw `ASSETS` and thought about building the assets. maybe then `INVENIO_JAVASCRIPT_PACKAGE_MANAGER_CLS` :) i know it is long, but we are using internally also that word for npm/pnpm

invenio_cli/helpers/package_managers.py
```diff
@@ -171,3 +174,57 @@ def start_activated_subshell(self) -> List[str]:
...
+        """Create a ``PNPMPackage`` with the path to ``package.json``."""
+        return PNPMPackage(package_json_path)
+
+    def install_local_package(self, path):
+        """Install the local JS package."""
+        return ["-C", str(path)]
+
+    def env_overrides(self):
+        """Provide environment overrides for building Invenio assets."""
+        return {"INVENIO_ASSETS_NPM_PKG_CLS": "pynpm:PNPMPackage"}
```
slint, line 230:
```suggestion
        return {"INVENIO_WEBPACKEXT_NPM_PKG_CLS": "pynpm:PNPMPackage"}
```

invenio_cli/helpers/package_managers.py
```diff
@@ -171,3 +178,186 @@ def start_activated_subshell(self) -> List[str]:
...
+            else:
+                return ProcessResponse(
+                    error=f"Unable to run postlink-dist. Got error code {status_code}",
+                    status_code=status_code,
+                )
+
+        return [
+            AssetsFunction(_prelink_dist, "Executing prelink-dist script..."),
+            AssetsFunction(_link_package_single_step, "Linking module to assets..."),
+            AssetsFunction(_postlink_dist, "Executing postlink-dist script..."),
```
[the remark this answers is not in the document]
max-moser: as discussed, we're basically executing the same scripts with `pnpm` as we did with `npm`, except for the part where the actual linking happens (which is more sane with `pnpm`.

since we don't have `preprelink-dist`, `postprelink-dist`, `prepostlink-dist` and `postpostlink-dist` scripts defined in our packages i think we should be executing the same set of scripts as before here.

invenio_cli/helpers/package_managers.py
```diff
@@ -171,3 +178,186 @@ def start_activated_subshell(self) -> List[str]:
...
+        ) -> ProcessResponse:
+            """Execute the PNPM single-step package linking."""
+            try:
+                # Accessing `package_json` fails if the file can't be found
+                module_pkg.package_json["name"]
+
+                # This is geared towards Invenio JS packages...
+                # But so are all the other steps
+                status_code = assets_pkg.link(
+                    str(module_pkg.package_json_path.parent / "dist")
```
max-moser, line 324: the fact that we're linking against the `"dist"` directory and not using the scripts requires some knowledge about the target package... but so does executing the `link-dist` script, which is not exactly a standard AFAIK :sweat_smile:

invenio_cli/helpers/package_managers.py
```diff
@@ -171,3 +178,186 @@ def start_activated_subshell(self) -> List[str]:
         # Also, it has a good chance of not properly setting a PS1...
         shell = os.getenv("SHELL")
         return [shell, "-c", f"source .venv/bin/activate; exec {shell} -i"]
+
+
+@dataclass
+class AssetsFunction:
+    """Function to run an assets command with `assets_pkg` and `module_pkg`."""
+
+    function: Callable[[NPMPackage, NPMPackage], ProcessResponse]
```
max-moser, line 187: the dataclass wants to have _some_ type annotation, but `Callable` should do the trick already.
i have no strong attachment to this type hint.

max-moser:
Once again (like with the PRs for Flask-WebpackExt & Invenio-Assets), we've discussed this feature and its implementation during the InvenioRDM community workshop 2025 etc. so the lack of official approval shouldn't indicate my rebellion here.

PR #386, opened 2025-03-01
versions: add support for version checks via pyproject.toml

max-moser:
This PR should've been part of #384 but was overlooked.
It adds support for parsing the `App-{RDM,ILS}` dependency version from `pyproject.toml`.


invenio_cli/helpers/versions.py
```diff
@@ -0,0 +1,74 @@
...
+#
+# Invenio is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""Invenio CLI dependencies helper."""
+
+import os
+import re
+
+import tomli
```
max-moser, line 15: `tomllib` is part of the standard lib starting with python 3.11, but until then we can use `tomli` which is a dependency from somewhere (`pipfile`?)

invenio_cli/helpers/versions.py
max-moser, line 1: merged `ils.py` and `rdm.py` into this one

invenio_cli/helpers/versions.py
slint: Maybe we add a direct dependency to `tomllib` with a Python version specifier? On my machine I get just a `toml` dependency when e.g. running on Python 3.9:

```
toml v0.10.2
└── pipfile v0.0.2 [requires: toml *]
    └── invenio-cli v1.6.0 [requires: pipfile >=0.0.2]
```

max-moser: right, i actually forgot that you can add dependencies for specific python versions only

max-moser: leaving this here for now: https://github.com/hukkin/tomli?tab=readme-ov-file#building-a-tomlitomllib-compatibility-layer

max-moser:
briefly discussed with @ntarocco :heavy_check_mark:

PR #387, opened 2025-03-01
rpc-server: use

utnapischtim:
* use rpc-server which runs invenio commands

NOTE:

maybe it is not a good idea to use the rpc-server on all `run_command` calls e.g. `invenio-cli assets watch`

### DIFF
--- a/invenio_cli/helpers/package_managers.py
+++ b/invenio_cli/helpers/package_managers.py
@@ -1,15 +1,20 @@
 # -*- coding: utf-8 -*-
 #
 # Copyright (C) 2025 TU Wien.
+# Copyright (C) 2025 Graz University of Technology.
 #
 # Invenio-Cli is free software; you can redistribute it and/or modify it
 # under the terms of the MIT License; see LICENSE file for more details.
 
 """Wrappers around various package managers to be used under the hood."""
 
+import atexit
 import os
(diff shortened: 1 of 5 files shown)


invenio_cli/helpers/package_managers.py
```diff
@@ -60,6 +87,24 @@ class Pipenv(PythonPackageManager):
 
     name = "pipenv"
     lock_file_name = "Pipfile.lock"
+    run_prefix = ["pipenv", "run"]
+
+    def send_command(self, *command):
+        """Send command to rpc server, default to run_command."""
```
ts-7bd, line 93: this function is untested

invenio_cli/helpers/package_managers.py
```diff
@@ -117,6 +162,25 @@ class UV(PythonPackageManager):
 
     name = "uv"
     lock_file_name = "uv.lock"
+    run_prefix = ["uv", "run", "--no-sync"]
+
+    def send_command(self, *command):
+        """Send command to rpc server, default to run_command."""
```
ts-7bd, line 168: this is almost a duplicate. maybe it is possible to reduce code with using super().send_command(..)

utnapischtim:
closed in favor of #390

PR #390, opened 2025-03-30
add rpc server lock

### LINKED ISSUE #179: Ensure JS reproducible builds
The `package-lock.json` file is not included in `my-site` folder, and builds are not fully reproducible, as the JS dependencies are always re-locked.

## Possible solution
* always copy `package-lock.json` from `../var/instance/assets/` to `my-site/`. Make sure that `git` ignores the creation/modification file dates if the file content didn't change
* have an extra CLI param in `invenio-cli` that will:
    * copy the `package-lock.json` from `my-site` to `../var/instance/assets/`
    * if the `my-site/pipfile.lock` is newer than `my-site/package-lock.json`, then overwrite or fail (some python deps might have changed)
    * run `npm ci` instead of `npm install` 
* the development workflow with linked editable (watched) modules should be tested, to ensure nothing breaks
* when in `NODE_ENV=production`, the above should always happen. When in `NODE_ENV=dev`, the I can use a new param `invenio-cli install --keep-deps`

utnapischtim:
- rpc-server: use  (needs https://github.com/inveniosoftware/invenio-app/pull/99) 
- fix: warning
- assets: add package lock file
- README: add new features

commit https://github.com/inveniosoftware/invenio-cli/pull/390/commits/100434ca4485c105b67ce98262ff89e863350d9b would close https://github.com/inveniosoftware/product-rdm/issues/179

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,6 +1,7 @@
 ..
     Copyright (C) 2019-2020 CERN.
     Copyright (C) 2019-2020 Northwestern University.
+    Copyright (C) 2025 Graz University of Technology.
 
     Invenio-Cli is free software; you can redistribute it and/or modify
     it under the terms of the MIT License; see LICENSE file for more details.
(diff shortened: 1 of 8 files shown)


README.rst
```diff
@@ -63,6 +64,54 @@ Local Development environment
...
+is a `uv.lock` file packages have to be updated manually or by removing the
+`uv.lock` file. The absence of the `uv.lock` file triggeres a new dependency
+resolving call which takes into account of new released packages. It would also
+be possible to use the `uv sync --upgrade` feature of `uv` but this installs
+also beta versions of packages which is not recommended. It may sound strange to
+remove the `uv.lock` file, but `uv` is that fast that deleting the `.venv`
+directory and the `uv.lock` file is the easiest and fastest and safest way to
+upgrade the packages.
+
+`rcp-server`
```
ts-7bd, line 110: kleiner typo

PR #391, opened 2025-03-31
Fix accidental creation of a tuple in version check

max-moser:
This fixes the `invenio-cli check-requirements` being broken when using `pipenv` (`Pipfile`).


invenio_cli/helpers/versions.py
```diff
@@ -35,7 +35,7 @@ def _parse_version(version):
 def _from_pipfile(dep_name):
     """Parse the stated dependency from the ``Pipfile``."""
     parsed = Pipfile.load(filename="./Pipfile")
-    version = (parsed.data["default"].get(dep_name, {}).get("version", ""),)
```
max-moser, line 38: the trailing comma created a tuple where we only want a string

utnapischtim (approved):

PR #392, opened 2025-04-02
feat(cli): default to pnpm, standardize services to postgresql and opensearch2

Samk13:

### Description

needs: https://github.com/inveniosoftware/cookiecutter-invenio-rdm/pull/304

To test, clone the repo checkout this branch, and install it with `uv pip install -e ~/yourpathto/invenio-cli` on your instance venv

### Checklist

- [ ] I'm aware of the [code of conduct](https://inveniordm.docs.cern.ch/contribute/code-of-conduct/).
- [ ] I've created [logical separate commits](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#commits) and followed the [commit message format](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#commit-message).
- [ ] I've added relevant test cases.
- [ ] I've added relevant documentation.
- [ ] I've marked [translation strings](https://inveniordm.docs.cern.ch/develop/howtos/i18n/).
- [ ] I've identified the [copyright holder(s)](https://inveniordm.docs.cern.ch/contribute/copyright-policy/) and updated copyright headers for touched files (>15 lines contributions).
- [ ] I've NOT included third-party code (copy/pasted source code or new dependencies).
    * If you have added [third-party code (copy/pasted or new dependencies)](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#third-party-codedependencies), please reach out to an [architect on Discord](https://discord.gg/8qatqBC).

**Frontend**

- [ ] I've followed the [CSS/JS](https://inveniordm.docs.cern.ch/develop/best-practices/css-js/) and [React](https://inveniordm.docs.cern.ch/develop/best-practices/react/) guidelines.
- [ ] I've followed the [web accessibility](https://inveniordm.docs.cern.ch/develop/best-practices/accessibility/) guidelines.
- [ ] I've followed the [user interface](https://inveniordm.docs.cern.ch/develop/best-practices/ui/) guidelines.

**Reminder**

### DIFF
--- a/invenio_cli/helpers/cli_config.py
+++ b/invenio_cli/helpers/cli_config.py
@@ -166,15 +166,13 @@ class CLIConfig(object):
         """Returns web host."""
         return self.private_config[CLIConfig.CLI_SECTION].get("web_host", "127.0.0.1")
 
-    @staticmethod
-    def get_db_type():
+    def get_db_type(self):
         """Returns the database type."""
-        return "postgresql"
+        return self.config[CLIConfig.COOKIECUTTER_SECTION].get("database", "postgresql")
 
-    @staticmethod
-    def get_search_type():
+    def get_search_type(self):
         """Returns the search type."""
-        return "opensearch2"
+        return self.config[CLIConfig.COOKIECUTTER_SECTION].get("search", "opensearch2")
 
     def get_file_storage(self):
         """Returns the file storage (local, s3, etc.)."""
@@ -223,8 +221,8 @@ class CLIConfig(object):
             config_parser[cls.COOKIECUTTER_SECTION][key] = str(value)
         # Keep compatibility with older tooling that expects `database` and `search` to exist.
         # Backend choice has been removed; PostgreSQL and OpenSearch2 are fixed.
-        config_parser[cls.COOKIECUTTER_SECTION]["database"] = cls.get_db_type()
-        config_parser[cls.COOKIECUTTER_SECTION]["search"] = cls.get_search_type()
+        config_parser[cls.COOKIECUTTER_SECTION]["database"] = "postgresql"
+        config_parser[cls.COOKIECUTTER_SECTION]["search"] = "opensearch2"
 
         # Generated files section
         config_parser[cls.FILES_SECTION] = get_created_files(project_dir)
--- a/invenio_cli/helpers/cli_config.py
+++ b/invenio_cli/helpers/cli_config.py
@@ -96,7 +96,7 @@ class CLIConfig(object):
         elif manager_name == PNPM.name:
             return PNPM()
 
-        return NPM()
+        return PNPM()
 
     def get_project_dir(self):
         """Returns path to project directory."""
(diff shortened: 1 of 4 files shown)


invenio_cli/helpers/cli_config.py
```diff
@@ -227,11 +228,17 @@ def write(cls, project_dir, flavour, replay):
         config_parser[cls.CLI_SECTION] = {}
         config_parser[cls.CLI_SECTION]["flavour"] = flavour
         config_parser[cls.CLI_SECTION]["logfile"] = "/logs/invenio-cli.log"
-
+        custom_package_managers = {
+            "python_package_manager": "uv",
+            "javascript_package_manager": "pnpm",
```
max-moser, line 233:
```suggestion
            "javascript_package_manager": replay.get("javascript_package_manager", None) or self.javascript_package_manager.name,
```
note: this is just from the top of my head, i doubt that it would work as is

from a quick check, it looks like this code path is only really executed in `invenio-cli init`, so accessing the replay would make a lot of sense i think
otherwise (if no replay is available... idk, would it be `None`?) we could fall back to the auto-detection logic above (which most likely defaults back to `npm` because we don't have any real detection logic there yet)

personally i would omit the `python_package_manager` as the auto-detection logic there should be fine unless you have both a `Pipfile` *and* `pyproject.toml` in your directory

Samk13: Resolved after a DM on Discord.

tmorrell (approved):
I think this is good to go, but we'll need a way to say this is for v14. Maybe this will start the invenio-cli v2.x series?

fenekku (approved):
Good by me (we'll do the WEBPACKEXT injection in a separate one)

invenio_cli/helpers/cli_config.py
```diff
@@ -165,24 +166,15 @@ def get_web_host(self):
         """Returns web host."""
         return self.private_config[CLIConfig.CLI_SECTION].get("web_host", "127.0.0.1")
 
-    def get_db_type(self):
-        """Returns the database type (mysql, postgresql)."""
-        return self.config[CLIConfig.COOKIECUTTER_SECTION]["database"]
```
fenekku, line 170: We don't lose anything by keeping this for now right? By keeping this logic that reads it from the file and the one below that writes it to the file, we keep backwards compatibility. This will reduce the corner cases for now.

fenekku (commented):
I'd say to keep backwards compatibilty as much as we can here for now.

invenio_cli/helpers/cli_config.py
Samk13: Since database and search are no longer [cookiecutter inputs](https://github.com/Samk13/cookiecutter-invenio-rdm/blob/860252f4f44840910513b63b60463e8e4403a6d6/cookiecutter.json), keeping the current implementation would fail when those keys are missing, so we'd need to get them with fallback values (postgresql/opensearch2) anyway.

Could you give an example of a backwards-compatibility scenario where we'd want to preserve a value other than PostgreSQL/OpenSearch2, given that the generated stack is now fixed to those services for `invenioRDM`?

fenekku: So here (before) "database" would be retrieved from the config file, so if we don't retrieve it from there and serve postgresql/opensearch2 directly this may contradict what someone had in that config file, right 🤔 ? Or did we change something else along the way? The context being someone using invenio-cli with an instance cookiecutted a couple version back.

I do see we are using get_db_type / get_search_type to fuel the writing now but not before, eh? I'll follow-up tomorrow 😄

Samk13: Updated this now anyway to keep compatibility and keep everyone happy.

The getters still read existing `.invenio` values, with fallbacks for new configs where cookiecutter no longer provides them.

The values in `write()` are hardcoded because it is a classmethod creating the config file, so there is no `self`/`CLIConfig` instance to read config from yet.

I will squash all commits once we agree.

fenekku: Yup, I think that's the way to go for this transition. At a later time we can remove even more.

fenekku (approved):
👍  Let's touch base at the telecon and get things merged after to make sure others are aligned.

utnapischtim (approved):

PR #399, opened 2025-07-17
local: add custom Jobs Celery Beat scheduler to `run`

slint:
> Disclaimer: this code was mainly generated using Claude Code.

- Adds **by default** running the jobs Celery beat custom scheduler as part of the `invenio-cli run` commands
- Adds a `--jobs-scheduler/--no-jobs-scheduler` flag

### DIFF
--- a/invenio_cli/commands/local.py
+++ b/invenio_cli/commands/local.py
@@ -201,7 +201,14 @@ class LocalCommands(Commands):
         """Run Celery beat scheduler for jobs."""
         # Jobs scheduler is only available in RDM v13+
         version = rdm_version()
-        if not version or version[0] < 13:
+        if version is None:
+            click.secho(
+                "RDM version couldn't be determined. Not running jobs scheduler.",
+                fg="yellow",
+                err=True,
+            )
+            return []
+        elif version[0] < 13:
             return []
 
         click.secho("Starting jobs scheduler...", fg="green")
--- a/invenio_cli/commands/services.py
+++ b/invenio_cli/commands/services.py
@@ -228,7 +228,14 @@ class ServicesCommands(Commands):
             steps.extend(self.fixtures())
             if demo_data:
                 steps.extend(self.demo())
-        elif ils_version():
+        else:
+            click.secho(
+                "RDM version couldn't be determined. RDM-specific steps will not be executed.",
+                fg="yellow",
+                err=True,
+            )
+
+        if ils_version():
             cmd = pkg_man.run_command("invenio", "setup", "--verbose")
             if not demo_data:
                 cmd.append("--skip-demo-data")


invenio_cli/commands/local.py
```diff
@@ -183,9 +185,43 @@ def run_worker(self, celery_log_file=None, celery_log_level="INFO"):
...
+            processes.extend(self.run_jobs_scheduler(celery_log_file, celery_log_level))
+
+        return processes
+
+    def run_jobs_scheduler(self, celery_log_file=None, celery_log_level="INFO"):
+        """Run Celery beat scheduler for jobs."""
+        click.secho("Starting jobs scheduler...", fg="green")
+
+        pkg_man = self.cli_config.python_package_manager
+        beat_command = pkg_man.run_command(
```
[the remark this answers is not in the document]
slint: Yes, they run in the same "call", similar to how `invenio-cli run` runs both worker and web processes. 

I considered originally adding it as a separate `invenio-cli run jobs-scheduler` sub-command, but I'm not sure this adds anything (since we're running anyways the default "beat" using the `--beat` flag on the main Celery worker command).

slint:
Maybe a caveat that I'm not 100% sure how to address is that `invenio-jobs` comes with v13 only... Should [we call `rdm_version()` as done in other places](https://github.com/inveniosoftware/invenio-cli/blob/master/invenio_cli/commands/services.py#L195) to "feature-detect" maybe?

utnapischtim:
> Maybe a caveat that I'm not 100% sure how to address is that `invenio-jobs` comes with v13 only... Should [we call `rdm_version()` as done in other places](https://github.com/inveniosoftware/invenio-cli/blob/master/invenio_cli/commands/services.py#L195) to "feature-detect" maybe?

yeah i would like to change to something else too. since we are not supporting the old versions anyway anymore.

and getting the version from the pyproject.toml file has also some drawbacks!

invenio_cli/commands/local.py
```diff
@@ -183,9 +186,47 @@ def run_worker(self, celery_log_file=None, celery_log_level="INFO"):
...
+        if jobs_scheduler:
+            processes.extend(self.run_jobs_scheduler(celery_log_file, celery_log_level))
+
+        return processes
+
+    def run_jobs_scheduler(self, celery_log_file=None, celery_log_level="INFO"):
+        """Run Celery beat scheduler for jobs."""
+        # Jobs scheduler is only available in RDM v13+
+        version = rdm_version()
+        if not version or version[0] < 13:
```
ntarocco, line 204: Nice!

ntarocco (approved):
It looks great, thanks a lot!

invenio_cli/commands/local.py
slint: I thought of a couple of alternatives:

- have this information coming [from `CLIConfig`](https://github.com/inveniosoftware/invenio-cli/blob/77e871406449717c4f86d7c47ca1eacc58953af1/invenio_cli/helpers/cli_config.py#L30), so that one e.g. could explicitly set something like `version = 13` in their `.invenio` and use the whole "Pipfile/pyproject.toml detection" dance as a fallback...
- get the version from the package manager (i.e. run `uv pip show invenio-app-rdm` or the `pipenv` equivalent), but I think that was the case in the past but is flakey since you might not have at all points the dependencies installed...
- get from the lockfiles. has similar caveats as the above point, where the lockfile might not have been generated yet

utnapischtim: > > i would prefer to crash directly, if version is not a valid value
> 
> Maybe you prefer to crash with a specific message like "Cannot detect app rdm version" or similar?

yes, that would be the best option. since the app-rdm version is a important information for invenio-cli, it should fail early in the usage of `invenio-cli`

utnapischtim: > I thought of a couple of alternatives:
> 
>     * have this information coming [from `CLIConfig`](https://github.com/inveniosoftware/invenio-cli/blob/77e871406449717c4f86d7c47ca1eacc58953af1/invenio_cli/helpers/cli_config.py#L30), so that one e.g. could explicitly set something like `version = 13` in their `.invenio` and use the whole "Pipfile/pyproject.toml detection" dance as a fallback...

that was my attempt to solve my problem. i added a value to the `.invenio.private` file and added a method `CLIConfig`

ntarocco: That would require that on each upgrade, we ask users to change the `.invenio` file and add/bump the version, right?

utnapischtim: no, that would mean, that users which don't add `invenio-app-rdm` into their `pyproject.toml` file of their instance, they have to change it in their .invenio.private file

which will not be that often, and if you customize your setup that much means you are, in my point of view, a power user, and those should have more options.

so my suggestions is not to replace the current way of doing it, only to add a new place where the app-rdm version could be picked up

PR #412, opened 2026-04-12
Add js lock file commands

tmorrell:

### Description

This PR builds off of some commits in https://github.com/inveniosoftware/invenio-cli/pull/390

At the moment, invenio-cli will wipe and recreate all js assets when runming the `invenio-cli install` command. This isn't the most safe thing to do in a world of package compromises. This PR attempts to recommend a new workflow, where a command `invenio-cli assets lock` will create a packages.json and package-lock.json`, which can be committed into a repository. Then the assets build commands use those locked dependencies when reinstalling. This makes the js behavior more similar to how the python behavior works (respecting a lock file), without changing the default if a lock file isn't present.

I haven't tested on pnpm, though support should be included.

### Checklist

- [x] I'm aware of the [code of conduct](https://inveniordm.docs.cern.ch/community/code-of-conduct/).
- [x] I've created [logical separate commits](https://inveniordm.docs.cern.ch/community/code/best-practices/commits/#commits) and followed the [commit message format](https://inveniordm.docs.cern.ch/community/code/best-practices/commits/#commit-message).
- [ ] I've added relevant test cases.
- [ ] I've added relevant documentation.
- [x] I've marked [translation strings](https://inveniordm.docs.cern.ch/community/translations/i18n/).
- [x] I've identified the [copyright holder(s)](https://inveniordm.docs.cern.ch/community/copyright-policy/) and updated copyright headers for touched files (>15 lines contributions).
- [x] I've NOT included third-party code (copy/pasted source code or new dependencies).
    * If you have added [third-party code (copy/pasted or new dependencies)](https://inveniordm.docs.cern.ch/community/code/best-practices/commits/#third-party-codedependencies), please reach out to an [architect on Discord](https://discord.gg/8qatqBC).

**Frontend**

- [ ] I've followed the [CSS/JS](https://inveniordm.docs.cern.ch/community/code/best-practices/css-js/) and [React](https://inveniordm.docs.cern.ch/community/code/best-practices/react/) guidelines.
- [ ] I've followed the [web accessibility](https://inveniordm.docs.cern.ch/community/code/best-practices/accessibility/) guidelines.
- [ ] I've followed the [user interface](https://inveniordm.docs.cern.ch/community/code/best-practices/ui/) guidelines.

**Reminder**

### DIFF
--- a/invenio_cli/commands/assets.py
+++ b/invenio_cli/commands/assets.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 #
+# Copyright (C) 2026 California Institute of Technology.
 # Copyright (C) 2020 CERN.
 #
 # Invenio-Cli is free software; you can redistribute it and/or modify it
@@ -8,6 +9,7 @@
 """Invenio module to ease the creation and management of applications."""
 
 from pathlib import Path
+from shutil import copyfile
 
 import click
 
@@ -74,6 +76,24 @@ class AssetsCommands(LocalCommands):
                 status_code=status_code,
             )
 
+    def _cache_js_lock_files(self):
+        """Cache js lock files."""
+        instance_path = self.cli_config.get_instance_path()
+        project_dir = self.cli_config.get_project_dir()
+
+        lock_file = self.cli_config.javascript_package_manager.lock_file_name
+        target_path = project_dir / lock_file
+        source_path = instance_path / "assets" / lock_file
+        if not source_path.is_symlink():
+            copyfile(source_path, target_path)
+
+        target_path = project_dir / "package.json"
+        source_path = instance_path / "assets" / "package.json"
+        if not source_path.is_symlink():
+            copyfile(source_path, target_path)
+
+        return ProcessResponse(output="Cached js lock files.", status_code=0)
+
     def watch_assets(self):
         """High-level command to watch assets for changes."""
         watch_cmd = self.cli_config.python_package_manager.run_command(
@@ -137,3 +157,35 @@ class AssetsCommands(LocalCommands):
         )
 
         return steps
+
+    def lock(self, debug=True, log_file=None):
+        """Lock assets."""
+        py_pkg_man = self.cli_config.python_package_manager
+        js_pkg_man = self.cli_config.javascript_package_manager
+
+        lock_arg = self.cli_config.javascript_package_manager.lock_dependencies()
+
+        ops = [py_pkg_man.run_command("invenio", "collect", "--verbose")]
+        ops.append(py_pkg_man.run_command("invenio", "webpack", "clean", "create"))
+        lock_args = ["invenio", "webpack", "install"] + lock_arg
+        ops.append(py_pkg_man.run_command(*lock_args))
+        ops.append(self._cache_js_lock_files)
+        messages = {
+            "build": "Locking assets...",
+        }
+
+        with env(FLASK_DEBUG="1" if debug else "0"):
+            for op in ops:
+                if callable(op):
+                    response = op()
+                else:
+                    if op[-1] in messages:
+                        click.secho(messages[op[-1]], fg="green")
+                    response = run_interactive(
+                        op,
+                        env={"PIPENV_VERBOSITY": "-1", **js_pkg_man.env_overrides()},
+                        log_file=log_file,
+                    )
+                if response.status_code != 0:
+                    break
+        return response
--- a/invenio_cli/commands/local.py
+++ b/invenio_cli/commands/local.py
@@ -11,8 +11,9 @@
 import os
 import signal
 from distutils.dir_util import copy_tree
-from os import environ
+from os import environ, symlink
 from pathlib import Path
+from shutil import copyfile
 from subprocess import Popen as popen
 
 import click
@@ -79,6 +80,41 @@ class LocalCommands(Commands):
             status_code=0,
         )
 
+    def _symlink_locked_file(self):
+        """Symlink locked file."""
+        instance_path = self.cli_config.get_instance_path()
+        project_dir = self.cli_config.get_project_dir()
+
+        if self.cli_config.javascript_package_manager.name == "npm":
+            lock_file = "packages-lock.json"
+        elif self.cli_config.javascript_package_manager.name == "pnpm":
+            lock_file = "pnpm-lock.yaml"
+
+        source_path = project_dir / lock_file
+        target_path = instance_path / "assets" / lock_file
+
+        if Path(source_path).exists():
+            symlink(source_path, target_path)
+
+        return ProcessResponse(output="Symlinked locked file.", status_code=0)
+
+    def _cache_locked_file(self):
+        """Cache locked file."""
+        instance_path = self.cli_config.get_instance_path()
+        project_dir = self.cli_config.get_project_dir()
+
+        if self.cli_config.javascript_package_manager.name == "npm":
+            lock_file = "packages-lock.json"
+        elif self.cli_config.javascript_package_manager.name == "pnpm":
+            lock_file = "pnpm-lock.yaml"
+
+        target_path = project_dir / lock_file
+        source_path = instance_path / "assets" / lock_file
+        if not source_path.is_symlink():
+            copyfile(source_path, target_path)
+
+        return ProcessResponse(output="Cached locked file.", status_code=0)
+
     def update_statics_and_assets(self, force, debug=False, log_file=None):
         """High-level command to update less/js/images/... files.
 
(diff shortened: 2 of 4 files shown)


tmorrell:
Closing and re-opening for CI

invenio_cli/helpers/cli_config.py
```diff
@@ -169,6 +169,10 @@ def get_web_host(self):
         """Returns web host."""
         return self.private_config[CLIConfig.CLI_SECTION].get("web_host", "127.0.0.1")
 
+    def get_app_rdm_version(self):
```
utnapischtim: nit: this is a separate feature and should be in a separate commit

invenio_cli/helpers/versions.py
```diff
@@ -58,9 +59,12 @@ def _from_pyproject_toml(dep_name):
     return _parse_version(v.version)
 
 
-def rdm_version():
+def rdm_version(cli_config):
     """Return the latest RDM version."""
-    if os.path.isfile("./Pipfile"):
+    if app_rdm_version := cli_config.get_app_rdm_version():
```
utnapischtim: nit: this should be combined with `get_app_rdm_version` in `cli_config.py` to a separate commit

invenio_cli/commands/local.py
```diff
@@ -87,15 +108,22 @@ def update_statics_and_assets(self, force, debug=False, log_file=None):
...
+        install_lock_arg = self.cli_config.javascript_package_manager.install_lock()
         ops = [py_pkg_man.run_command("invenio", "collect", "--verbose")]
 
         if force:
             ops.append(py_pkg_man.run_command("invenio", "webpack", "clean", "create"))
-            ops.append(py_pkg_man.run_command("invenio", "webpack", "install"))
+            ops.append(self._symlink_locked_file)
+            ops.append(
+                py_pkg_man.run_command(
+                    "invenio", "webpack", "install", install_lock_arg
```
utnapischtim: this is not necessary. if `package.json` and lock file `pnpm-lock.yml` or `package-lock.json` is symlinked to the `assets` directory it is used automatically. adding an explicit command is not necessary, at least it wasn't since i am using this feature

invenio_cli/helpers/cli_config.py
tmorrell: It was in your commit :) I've removed this since although it's a really great simplification, it's not related to the package-lock stuff. You can drop it in your own PR.

invenio_cli/commands/local.py
tmorrell: The command is needed to ensure that the package-lock file is fully respected. https://stackoverflow.com/questions/73968943/how-to-have-pnpm-install-install-everything-exactly-to-the-specs-of-the-pnpm-l 

Otherwise the package-lock file can be modified during the install, and I ran into that in my testing. I was using a modified package-lock file to test whether things were getting saved correctly. In normal operation you probably wouldn't have run into it, but if we're saying we're installing from the package-lock I think we might as well really install from the package-lock

invenio_cli/helpers/cli_config.py
utnapischtim: hehe, yes it was in my commit but it was a WIP commit where i added to much in one commit. i planned to separate it afterwards.

i will add it in my PR :)

invenio_cli/commands/local.py
utnapischtim: hm, i see your point but i am still irritated. when do `RUN pnpm install` in my `Dockerfile` i see `Lockfile is up to date, resolution step is skipped` and thought that this means that further parameters are not necessary to enforce the usage of the lock file

as i understand the stackoverflow answer the lockfile will be changed if the `package.json` file has been changed which is the normal behavior. that's the reason why i linked the `package.json` file into the assets directory too. otherwise the package.json file would be always newer as the lockfile and therefore the lockfile would be recalculated. 

so adding the additional parameter would prevent that package.json file and the lockfile can be out of sync which is good but should never happen or?

tmorrell: I'm trying to get around how the `package.json` file is regenerated from the individual invenio packages. So if someone runs `invenio-cli install` the `package.json` file will change to whatever the latest state is. And because it's simlinked, the `package.json` file in the project directory will be updated.

So if we don't use the lock flag, the `package-lock.json` file will see the changed `package.json` file and then get updated before the install. This effectively gets around the locking.

Now if there are incompatibilities between `packages.json` and `package-lock.json` someone might have to run `invenio-cli assets build` in order to sort out those incompatibilities before the install will work....but to me that's kind of like deleting a pipfile.lock if you're trying to upgrade stuff.

There might be a better way to re-design the assets building, but I want to make sure when we say something is locked it's really locked.

utnapischtim: my idea was to let the `package.json` be created and then it is replaced by the `package.json` and `pnpm-lock.yaml` file which is cached in the root directory of the instance. which means that at the point `invenio webpack install` is triggered the symlink function has to be executed before which should be the case [here](https://github.com/utnapischtim/invenio-cli/blob/100434ca4485c105b67ce98262ff89e863350d9b/invenio_cli/commands/local.py#L129)

i think it is not possible to not generate the package.json file first at least for `invenio-cli`

and if there are incompatibilities they should be found with `invenio-cli install` too. as i remember `invenio-cli install` does use the same function as `invenio-cli assets build` uses.

> but to me that's kind of like deleting a pipfile.lock if you're trying to upgrade stuff.

yes that is was i normally do. i delete `uv.lock`, `package.json` and `pnpm-lock.yaml` when i want to have a fully updated instance. 

in my opinion as i encountered in the beginning the parameters are not necessary, but it could be that i overlooked a problem. so in doubt i am in favor of those parameters

tmorrell: Maybe we do a copy instead of a simlink? 

At the moment the first time you do the install it will simlink, and then the simlink will remain forever after. So your temporary `package.json` will overwrite whatever you had in your project directory.

I think with a copy you'd be able to ensure that the project directory version is what's used in the install, and then I could drop the extra parameter

invenio_cli/helpers/package_managers.py
```diff
@@ -223,6 +233,14 @@ def install_local_package(self, path):
         """Install the local JS package."""
         return ["--prefix", str(path)]
 
+    def lock_dependencies(self):
+        """Lock the JS package."""
+        return "--package-lock-only --ignore-scripts"
```
mirekys: should probably be passed as list of args rather than a single str arg

invenio_cli/commands/local.py
```diff
@@ -79,6 +81,25 @@ def _statics(self):
...
+    def _symlink_locked_file(self):
+        """Symlink locked file."""
+        instance_path = self.cli_config.get_instance_path()
+        project_dir = self.cli_config.get_project_dir()
+
+        lock_file = self.cli_config.javascript_package_manager.lock_file_name
+        source_path = project_dir / lock_file
+        target_path = instance_path / "assets" / lock_file
+        if Path(source_path).exists():
+            symlink(source_path, target_path)
```
mirekys: What if already symlinked to target path (instance assets)? It would throw FileExists here

invenio_cli/commands/local.py
```diff
@@ -79,6 +81,25 @@ def _statics(self):
...
+
+        source_path = project_dir / "package.json"
```
mirekys, line 95: I think responsibility of the function should be lockfile symlinking only. Package.json linking should have its own, unless it is crucial to keep it coupled here.

invenio_cli/commands/local.py
tmorrell: I've gone ahead and switched to the copy version. I think that makes it clearer what is going on, and then we don't need the extra parameter. Let me know what you think!

invenio_cli/commands/local.py
tmorrell: I've switched to a copy that I think makes the workflow clearer.

invenio_cli/commands/local.py
tmorrell: Package.json and package-lock.json are really strongly linked. If you don't have matching files, npm will just update the package-lock file to match. So I think we need to do both.

I have updated the function name to make that clearer.

invenio_cli/helpers/package_managers.py
tmorrell: That's more consistent. Changed!

invenio_cli/commands/local.py
tmorrell: I'm not fully convinced that this really respects the lock file without the parameter.

invenio_cli/commands/assets.py
```diff
@@ -137,3 +157,36 @@ def watch_js_module(self, path, link=True):
...
+        py_pkg_man = self.cli_config.python_package_manager
+        js_pkg_man = self.cli_config.javascript_package_manager
+
+        lock_arg = self.cli_config.javascript_package_manager.lock_dependencies()
+
+        ops = [py_pkg_man.run_command("invenio", "collect", "--verbose")]
+        ops.append(py_pkg_man.run_command("invenio", "webpack", "clean", "create"))
+        lock_args = ["invenio", "webpack", "install"] + lock_arg
+        ops.append(py_pkg_man.run_command(*lock_args))
+        ops.append(self._statics)
```
mirekys: Is this step required (statics shouldn't affect lockfile? - problably room for speedup?)

tmorrell: You're right, I've removed it.

invenio_cli/commands/local.py
tmorrell: The reason I was confused is that npm re-writes the lockfile, even if it respects it and doesn't change things. I think this approach is working fine.

tmorrell:
I think this is now fully ready and I've addressed all comments. If someone wants to approve I can get this out.

slint (approved):

PR #422, opened 2026-05-06
:package: release: v1.11.0

fenekku:


CHANGES.rst
```diff
@@ -10,6 +10,13 @@
 Changes
 =======
 
+Version v1.11.0 (released 2026-05-06)
+
+- fix(review): condition inverted
```
utnapischtim: nit: it would be enough to only state that teh feature is introduced. the commit which applies review comments does not have to be listed in the release notes

utnapischtim (approved):

CHANGES.rst
fenekku: I started automating my generation of these changelog sections. I am just doing a really simplistic "get a list of commits since last tag" right now, but I am curious what if any others are using (git-cliff, conventional-changelog-cli, cocogitto and co all seemed too much effort).

